PR #9, opened 2022-06-13
8287971: Throw exception for missing values in .jpackage.xml

sashamatveev:
- Error will be thrown if app image is generated with another JDK version or has malformed .jpackage.xml.
 - Re-fixed as Alexey suggested in https://github.com/openjdk/jdk/pull/9098.


src/jdk.jpackage/share/classes/jdk/jpackage/internal/resources/MainResources_de.properties
```diff
@@ -75,8 +75,9 @@ error.blocked.option=jlink-Option [{0}] ist in --jlink-options nicht zul\u00E4ss
 error.no.name=Name nicht mit --name angegeben. Es kann auch kein Name aus app-image abgeleitet werden
 
 warning.no.jdk.modules.found=Warnung: Keine JDK-Module gefunden
-warning.foreign-app-image=Warnung: app-image-Verzeichnis ({0}) wurde von jpackage nicht generiert.
-warning.invalid-app-image=Warnung: .jpackage.xml kann in app-image-Verzeichnis ({0}) nicht geparst werden
+
+error.foreign-app-image=Error: app-image dir ({0}) not generated by jpackage. Missing .jpackage.xml.
```
alexeysemenyukoracle, line 79: This error message will be misleading in case app image was generated by jpackage and adjusted after.
I'd put it as "Missing .jpackage.xml file in app-image dir ({0})."

test/jdk/tools/jpackage/helpers/jdk/jpackage/test/TKit.java
```diff
@@ -233,6 +233,43 @@ public static void createTextFile(Path filename, Stream<String> lines) {
...
+                                mainLauncher),
+                            String.format(
+                                "<main-class>%s</main-class>",
+                                mainClass),
+                            "<signed>false</signed>",
+                            "<app-store>false</app-store>",
+                            "</jpackage-state>"
+            ));
+    }
+
```
alexeysemenyukoracle, line 272: 1. This is jpackage specific, so it should belong to JPackageCommand class.
2. Please use proper XML API to create XML output.
3. Having `<app-version>1.0</app-version>` in .jpackage.xml file will make `AppImageFile.isValid()` return `false` and `AppImageFile.load()` to throw exception. Which seems to be equivalent of not having .jpackage.xml file in app image at all making it a question if this new function is need at all.

src/jdk.jpackage/share/classes/jdk/jpackage/internal/AppImageFile.java
```diff
@@ -332,29 +298,43 @@ private static String xpathQueryNullable(XPath xPath, String xpathExpr,
...
+        if (signedStr == null ||
+                !("true".equals(signedStr) || "false".equals(signedStr))) {
             return false;
         }
 
-        if (!Objects.equals(getPlatform(), creatorPlatform)) {
+        if (appStoreStr == null ||
+                !("true".equals(appStoreStr) || "false".equals(appStoreStr))) {
             return false;
         }
```
alexeysemenyukoracle, line 339: It makes sense to unfold this function in the ctor as we don't allow empty AppImageFile instances.

src/jdk.jpackage/share/classes/jdk/jpackage/internal/AppImageFile.java
```diff
@@ -64,20 +64,15 @@ public final class AppImageFile {
     private final String launcherName;
     private final String mainClass;
     private final List<LauncherInfo> addLauncherInfos;
-    private final boolean signed;
-    private final boolean appStore;
+    private final String signedStr;
+    private final String appStoreStr;
```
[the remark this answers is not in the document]
sashamatveev: To store string values, so it can be validated to make sure it is true or false only. I changed it back, since I moved validation to ctor.

src/jdk.jpackage/share/classes/jdk/jpackage/internal/AppImageFile.java
sashamatveev: Done.

src/jdk.jpackage/share/classes/jdk/jpackage/internal/resources/MainResources_de.properties
sashamatveev: Done.

test/jdk/tools/jpackage/helpers/jdk/jpackage/test/TKit.java
sashamatveev: 1 and 2 done. I think you mistaken `<app-version>1.0</app-version>` with `<jpackage-state version=\"%s\"`. We never read <app-version> and I do put correct value for `<jpackage-state version=\"%s\"`. It is required for several tests which uses empty/dummy app images and since we started required .jpackage.xml file we need it for empty app images as well.

alexeysemenyukoracle (changes requested):
No need to copy/paste xml authoring code from IOUtils. This API is accessible outside of jpackage. Check out https://github.com/openjdk/jdk/blob/739769c8fc4b496f08a92225a12d07414537b6c0/test/jdk/tools/jpackage/windows/WinScriptTest.java#L151

test/jdk/tools/jpackage/helpers/jdk/jpackage/test/TKit.java
alexeysemenyukoracle: Right! My bad, sorry for the confusion.

sashamatveev:
8287971: Throw exception for missing values in .jpackage.xml [v3]
 - Removed copy/paste code from TKit.java.

test/jdk/tools/jpackage/helpers/jdk/jpackage/test/JPackageCommand.java
```diff
@@ -298,6 +299,57 @@ public JPackageCommand setFakeRuntime() {
...
+            platform = "linux";
+        } else if (TKit.isOSX()) {
+            platform = "macOS";
+        } else {
+            platform = "unknown";
+        }
+
+        ApplicationLayout layout = ApplicationLayout.platformAppImage();
+        Path jpackageXMLFile = layout.resolveAt(Path.of(appImage))
+                .appDirectory().resolve(".jpackage.xml");
```
alexeysemenyukoracle, line 324: Can be simplified:
`Path jpackageXMLFile = AppImageFile.getPathInAppImage(appImage);`

test/jdk/tools/jpackage/helpers/jdk/jpackage/test/JPackageCommand.java
```diff
@@ -298,6 +299,57 @@ public JPackageCommand setFakeRuntime() {
         return this;
     }
 
+    public void createJPackageXMLFile(String mainLauncher, String mainClass)
+            throws IOException {
+        String appImage = getArgumentValue("--app-image");
+        if (appImage == null) {
+            throw new RuntimeException("Error: --app-image expected");
+        }
```
alexeysemenyukoracle, line 307: Can be simplified to:
```
Path jpackageXMLFile = AppImageFile.getPathInAppImage(
                Optional.ofNullable(getArgumentValue("--app-image")).map(
                        Path::of).orElseThrow(() -> {
                            return new RuntimeException(
                                    "Error: --app-image expected");
                        }));
```

And you don't need explicit `ApplicationLayout.platformAppImage()` call below.

alexeysemenyukoracle (changes requested):

test/jdk/tools/jpackage/helpers/jdk/jpackage/test/JPackageCommand.java
sashamatveev: Done. I also switched to use AppImageFile.getVersion() and AppImageFile.getPlatform().

alexeysemenyukoracle (approved):

PR #20, opened 2022-06-15
8288497: add support for JavaThread::is_oop_safe()

dcubed-ojdk:
A fix to add support for JavaThread::is_gc_barrier_detached() which allows
us to add checks to detect failures like:

    JDK-8288139 JavaThread touches oop after GC barrier is detached
    https://bugs.openjdk.org/browse/JDK-8288139

This fix along with the fix for JDK-8288139 has been tested in Mach5 Tier[1-8].
There are no related failures in Mach5 Tier[1-7]; Mach5 Tier8 is still running.

---------
### Progress

### Issue
 * [JDK-8288497](https://bugs.openjdk.org/browse/JDK-8288497): add support for JavaThread::is_oop_safe()

### Reviewers
 * [Patricio Chilano Mateo](https://openjdk.org/census#pchilanomate) (@pchilano - **Reviewer**)
 * [Robbin Ehn](https://openjdk.org/census#rehn) (@robehn - **Reviewer**) ⚠️ Review applies to [24ecdea4](https://git.openjdk.org/jdk19/pull/20/files/24ecdea4f0111057faf805c89d6bf1033f27b707)
 * [Erik Österlund](https://openjdk.org/census#eosterlund) (@fisk - **Reviewer**) ⚠️ Review applies to [24ecdea4](https://git.openjdk.org/jdk19/pull/20/files/24ecdea4f0111057faf805c89d6bf1033f27b707)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/20/head:pull/20` \
`$ git checkout pull/20`

`$ git checkout pull/20` \
`$ git pull https://git.openjdk.org/jdk19 pull/20/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 20`

`$ git pr show -t 20`

</details>

<a href="https://git.openjdk.org/jdk19/pull/20.diff">https://git.openjdk.org/jdk19/pull/20.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/thread.hpp
+++ b/src/hotspot/share/runtime/thread.hpp
@@ -904,8 +904,9 @@ class JavaThread: public Thread {
   // JavaThread termination support
  public:
   enum TerminatedTypes {
-    _not_terminated = 0xDEAD - 2,
+    _not_terminated = 0xDEAD - 3,
     _thread_exiting,                             // JavaThread::exit() has been called for this thread
+    _thread_gc_barrier_detached,                 // thread's GC barrier has been detached
     _thread_terminated,                          // JavaThread is removed from thread list
     _vm_exited                                   // JavaThread is still executing native code, but VM is terminated
                                                  // only VM_Exit can set _vm_exited
@@ -914,7 +915,7 @@ class JavaThread: public Thread {
  private:
   // In general a JavaThread's _terminated field transitions as follows:
   //
-  //   _not_terminated => _thread_exiting => _thread_terminated
+  //   _not_terminated => _thread_exiting => _thread_gc_barrier_detached => _thread_terminated
   //
   // _vm_exited is a special value to cover the case of a JavaThread
   // executing native code after the VM itself is terminated.
@@ -1172,13 +1173,17 @@ private:
   bool on_thread_list() const { return _on_thread_list; }
   void set_on_thread_list() { _on_thread_list = true; }
 
-  // thread has called JavaThread::exit() or is terminated
+  // thread has called JavaThread::exit(), thread's GC barrier is detached
+  // or thread is terminated
   bool is_exiting() const;
+  // thread's GC barrier is detached or thread is terminated
+  bool is_gc_barrier_detached() const;
   // thread is terminated (no longer on the threads list); we compare
-  // against the two non-terminated values so that a freed JavaThread
+  // against the three non-terminated values so that a freed JavaThread
   // will also be considered terminated.
   bool check_is_terminated(TerminatedTypes l_terminated) const {
-    return l_terminated != _not_terminated && l_terminated != _thread_exiting;
+    return l_terminated != _not_terminated && l_terminated != _thread_exiting &&
+           l_terminated != _thread_gc_barrier_detached;
   }
   bool is_terminated() const;
   void set_terminated(TerminatedTypes t);
--- a/src/hotspot/share/runtime/thread.hpp
+++ b/src/hotspot/share/runtime/thread.hpp
@@ -1180,8 +1180,8 @@ private:
   // thread has called JavaThread::exit(), thread's GC barrier is detached
   // or thread is terminated
   bool is_exiting() const;
-  // thread's GC barrier is detached or thread is terminated
-  bool cannot_access_oops_safely() const;
+  // thread's GC barrier is NOT detached and thread is NOT terminated
+  bool is_oop_safe() const;
   // thread is terminated (no longer on the threads list); we compare
   // against the three non-terminated values so that a freed JavaThread
   // will also be considered terminated.
--- a/src/hotspot/share/runtime/thread.hpp
+++ b/src/hotspot/share/runtime/thread.hpp
@@ -919,6 +919,10 @@ class JavaThread: public Thread {
   //
   // _vm_exited is a special value to cover the case of a JavaThread
   // executing native code after the VM itself is terminated.
+  //
+  // A JavaThread that fails to JNI attach has these _terminated field transitions:
+  //   _not_terminated => _thread_terminated
+  //
   volatile TerminatedTypes _terminated;
 
   jint                  _in_deopt_handler;       // count of deoptimization
@@ -1177,7 +1181,7 @@ private:
   // or thread is terminated
   bool is_exiting() const;
   // thread's GC barrier is detached or thread is terminated
-  bool is_gc_barrier_detached() const;
+  bool cannot_access_oops_safely() const;
   // thread is terminated (no longer on the threads list); we compare
   // against the three non-terminated values so that a freed JavaThread
   // will also be considered terminated.
--- a/src/hotspot/share/runtime/thread.inline.hpp
+++ b/src/hotspot/share/runtime/thread.inline.hpp
@@ -266,10 +266,10 @@ inline bool JavaThread::is_exiting() const {
          check_is_terminated(l_terminated);
 }
 
-inline bool JavaThread::cannot_access_oops_safely() const {
+inline bool JavaThread::is_oop_safe() const {
   TerminatedTypes l_terminated = Atomic::load_acquire(&_terminated);
-  return l_terminated == _thread_gc_barrier_detached ||
-         check_is_terminated(l_terminated);
+  return l_terminated != _thread_gc_barrier_detached &&
+         !check_is_terminated(l_terminated);
 }
 
 inline bool JavaThread::is_terminated() const {
--- a/src/hotspot/share/runtime/thread.inline.hpp
+++ b/src/hotspot/share/runtime/thread.inline.hpp
@@ -261,14 +261,24 @@ inline void JavaThread::set_done_attaching_via_jni() {
 
 inline bool JavaThread::is_exiting() const {
   // Use load-acquire so that setting of _terminated by
-  // JavaThread::exit() is seen more quickly.
+  // JavaThread::set_terminated() is seen more quickly.
   TerminatedTypes l_terminated = Atomic::load_acquire(&_terminated);
-  return l_terminated == _thread_exiting || check_is_terminated(l_terminated);
+  return l_terminated == _thread_exiting ||
+         l_terminated == _thread_gc_barrier_detached ||
+         check_is_terminated(l_terminated);
+}
+
+inline bool JavaThread::is_gc_barrier_detached() const {
+  // Use load-acquire so that setting of _terminated by
+  // JavaThread::set_terminated() is seen more quickly.
+  TerminatedTypes l_terminated = Atomic::load_acquire(&_terminated);
+  return l_terminated == _thread_gc_barrier_detached ||
+         check_is_terminated(l_terminated);
 }
 
 inline bool JavaThread::is_terminated() const {
   // Use load-acquire so that setting of _terminated by
-  // JavaThread::exit() is seen more quickly.
+  // JavaThread::set_terminated() is seen more quickly.
   TerminatedTypes l_terminated = Atomic::load_acquire(&_terminated);
   return check_is_terminated(l_terminated);
 }
--- a/src/hotspot/share/runtime/thread.inline.hpp
+++ b/src/hotspot/share/runtime/thread.inline.hpp
@@ -260,31 +260,24 @@ inline void JavaThread::set_done_attaching_via_jni() {
 }
 
 inline bool JavaThread::is_exiting() const {
-  // Use load-acquire so that setting of _terminated by
-  // JavaThread::set_terminated() is seen more quickly.
   TerminatedTypes l_terminated = Atomic::load_acquire(&_terminated);
   return l_terminated == _thread_exiting ||
          l_terminated == _thread_gc_barrier_detached ||
          check_is_terminated(l_terminated);
 }
 
-inline bool JavaThread::is_gc_barrier_detached() const {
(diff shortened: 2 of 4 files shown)


src/hotspot/share/runtime/thread.inline.hpp
```diff
@@ -261,14 +261,24 @@ inline void JavaThread::set_done_attaching_via_jni() {
 
 inline bool JavaThread::is_exiting() const {
   // Use load-acquire so that setting of _terminated by
-  // JavaThread::exit() is seen more quickly.
+  // JavaThread::set_terminated() is seen more quickly.
```
dcubed-ojdk, line 264: This comment should have been updated when the code in
JavaThread::exit() was refactored into JavaThread::set_terminated().
I'm doing it as part of this fix for clarity.

src/hotspot/share/runtime/thread.inline.hpp
```diff
@@ -261,14 +261,24 @@ inline void JavaThread::set_done_attaching_via_jni() {
...
+  TerminatedTypes l_terminated = Atomic::load_acquire(&_terminated);
+  return l_terminated == _thread_gc_barrier_detached ||
+         check_is_terminated(l_terminated);
 }
 
 inline bool JavaThread::is_terminated() const {
   // Use load-acquire so that setting of _terminated by
-  // JavaThread::exit() is seen more quickly.
+  // JavaThread::set_terminated() is seen more quickly.
```
dcubed-ojdk, line 281: This comment should have been updated when the code in
JavaThread::exit() was refactored into JavaThread::set_terminated().
I'm doing it as part of this fix for clarity.

src/hotspot/share/services/threadService.cpp
```diff
@@ -163,7 +163,8 @@ void ThreadService::remove_thread(JavaThread* thread, bool daemon) {
 
   assert(!thread->is_terminated(), "must not be terminated");
   if (!thread->is_exiting()) {
-    // JavaThread::exit() skipped calling current_thread_exiting()
+    // We did not get here via JavaThread::exit() so current_thread_exiting()
+    // was not called, e.g., JavaThread::cleanup_failed_attach_current_thread().
```
dcubed-ojdk, line 167: [JDK-8286830](https://bugs.openjdk.org/browse/JDK-8286830) ~HandshakeState should not touch oops
changed JavaThread::exit() so we no longer had a path that called current_thread_exiting()
and another path that did not call current_thread_exiting() so this comment should have
been clarified with that fix. I'm doing it as part of this fix for clarity.

pchilano (commented):
I think I would just use the _thread_terminated state rather than introducing a new _thread_gc_barrier_detached state just to detect failures. So the only extra things covered by _thread_gc_barrier_detached and not by _thread_terminated would be ThreadService::remove_thread() which just updates some counters, and ThreadsSMRSupport::remove_thread() which we overlooked before but now by inspection I don't see any other place where we touch oops there.

dcubed-ojdk:
@pchilano - Thanks for the review!

Since we've been having issues with oop usage after the GC barrier is detached
in several bugs, I was going for the more precise control of adding the new
TerminatedTypes value. There is quite a bit of code from:

old L3600: `BarrierSet::barrier_set()->on_thread_detach(p);`

to

old L3623: `p->set_terminated(JavaThread::_thread_terminated);` 

so rather than have to reason about the code over and over again,
I would rather do these sanity checks with targeted code. To paraphrase
@robehn: Why do with code inspection what you can do with code?

dcubed-ojdk:
If the folks that own the barrier code add detection checks there, then this code
can be removed in the future. For now, I would like to get this code in so that I
can continue to search for additional instances of this type of failure.

As for the addition of the _thread_gc_barrier_detached TerminatedTypes value
and the location of when that state goes active, it is that new state and it's
placement that verified that the code in ThreadsSMRSupport::remove_thread()
was in the wrong place (fixed in JDK-8288139). Depending on the
_thread_terminated TerminatedTypes value did not and would not have verified
that problem.

dholmes-ora (commented):
Hi Dan,

After looking at the barrier code I'm warming up to the proposal you have made, but with a couple of suggestions below.

Thanks.

fisk:
I have also been bitten before by oops being touched at the boundaries of the thread life cycle, so I understand why we would want to be more precise about modelling this in the thread life cycle. The only thing I don't understand with the patch is why there are two alternative state machines now. One that goes through the new state and one that does not go through the new state. There might be a good reason for that, but I would like to understand why that is. And probably have that written in a comment where the two state machines are described.

robehn:
Hi,

Except for the few lines of code between:
BarrierSet::barrier_set()->on_thread_detach(p) and ThreadsSMRSupport::remove();

This new state is identical to the result of "ThreadsSMRSupport::get_java_thread_list()->includes(p)".
The JavaThread is attached to the barrier set when we add it to the ThreadsList and detached when we remove it.

If we have a safepoint poll between (done with Threads_lock held):
BarrierSet::barrier_set()->on_thread_detach(p);
->
p->set_terminated(JavaThread::_thread_terminated);

We have a bug. This means that this new state is not even possible to observe from another thread (with target safe).
So for any other thread there is no difference, which means we have new global state just for a few lines of code while we are terminating with Threads_lock held.

So I don't think this one place needs a new global state.

Note that for a safe target thread the terminated flag gives the same result as:
"ThreadsSMRSupport::get_java_thread_list()->includes(p)"

So jt->is_terminated() can be implemented with "ThreadsSMRSupport::get_java_thread_list()->includes(p)". (again assuming we are not doing some racy, and only asking for threads that are safe/stable)

zhengyu123:
Just a suggestion:  I believe adding thread state check in Thread::gc_data() method helps to catch this problem in the future, as the gc_data is usually accessed by GC barrier ...

src/hotspot/share/runtime/thread.hpp
```diff
@@ -1172,13 +1173,17 @@ class JavaThread: public Thread {
   bool on_thread_list() const { return _on_thread_list; }
   void set_on_thread_list() { _on_thread_list = true; }
 
-  // thread has called JavaThread::exit() or is terminated
+  // thread has called JavaThread::exit(), thread's GC barrier is detached
+  // or thread is terminated
   bool is_exiting() const;
+  // thread's GC barrier is detached or thread is terminated
+  bool is_gc_barrier_detached() const;
```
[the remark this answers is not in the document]
dcubed-ojdk: We can definitely rename, but naming is hard... :-)

Those names you proposed are inversions of the current name and would
require that the logic of the function be inverted.

The other names I came up before I settled on is_gc_barrier_detached():

```
    can_no_longer_access_oops()
    oops_are_no_longer_safe()
    cannot_safely_access_oops()
```

The closest match between your proposed names and mine are:

`can_access_oops_safely`  and `cannot_safely_access_oops`

Other opinions are welcome.

src/hotspot/share/runtime/thread.hpp
```diff
@@ -1172,13 +1173,17 @@ class JavaThread: public Thread {
...
   // thread is terminated (no longer on the threads list); we compare
-  // against the two non-terminated values so that a freed JavaThread
+  // against the three non-terminated values so that a freed JavaThread
   // will also be considered terminated.
   bool check_is_terminated(TerminatedTypes l_terminated) const {
-    return l_terminated != _not_terminated && l_terminated != _thread_exiting;
+    return l_terminated != _not_terminated && l_terminated != _thread_exiting &&
+           l_terminated != _thread_gc_barrier_detached;
```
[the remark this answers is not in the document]
dcubed-ojdk: The logic style of this function is a carry over from before we had Thread-SMR
and we tried to detect a freed JavaThread. With Thread-SMR, we don't need to
do the check this way at all.

Do we really want to change that as part of this fix for JDK19?

src/hotspot/share/runtime/thread.inline.hpp
dcubed-ojdk: I'm pretty sure that I wrote the original comment and you were one of my
reviewers way, way back then.

When we were trying to match up the load-acquires in the accessors with
the in-line release-store code in JavaThread::exit(), the comment was needed
(ignoring the part about "seeing things more quickly"). Now that we have
load-acquires in the accessors and a release-store in the setter, i.e., no more
in-line code in JavaThread::exit(), I no longer think we need a comment to
coordinate.

I'm in favor of dropping all instances of this comment.

src/hotspot/share/runtime/thread.inline.hpp
dcubed-ojdk: See above discussion about this comment.
I'm in favor of dropping this comment.

dcubed-ojdk:
@fisk's comment from above:
> The only thing I don't understand with the patch is why there are two alternative
> state machines now. One that goes through the new state and one that does not
> go through the new state. There might be a good reason for that, but I would like
> to understand why that is. And probably have that written in a comment where
> the two state machines are described.

There used to be (at least) three alternative state machines. There were two in
JavaThread::exit() and the failing JNI attach is the third. There might be more,
but I haven't found them yet.

dcubed-ojdk:
@zhengyu123's comment from above:

> Just a suggestion: I believe adding thread state check in Thread::gc_data()
> method helps to catch this problem in the future, as the gc_data is usually
> accessed by GC barrier ...

I think we (Runtime) are hoping that the GC folks add something specific that
is intended to catch these kind of problems. If that happens, then this hack
that I'm doing now can be retired. Adding a "thread state" check to
Thread::gc_data() might be the solution, but I would feel more comfortable if
that change was made by GC folks that know this barrier stuff.

src/hotspot/share/runtime/thread.hpp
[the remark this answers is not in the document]
dholmes-ora: I think the style guide for naming is to use positive names to avoid double negatives when you have to negate the function - so `is_safe_for_x` rather than `is_not_safe_for_x`. I like `thread->is_oop_safe()` - short and clear.

src/hotspot/share/runtime/thread.hpp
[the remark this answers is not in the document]
dholmes-ora: No it can be cleaned up later. This fix just highlighted the badness of that piece of code. :)

robehn:
> @robehn's comment from above:
> 
> > Except for the few lines of code between:
> > BarrierSet::barrier_set()->on_thread_detach(p) and ThreadsSMRSupport::remove();
> > This new state is identical to the result of "ThreadsSMRSupport::get_java_thread_list()->includes(p)".
> 
> No it's not equivalent. If I replace is_gc_barrier_attached()'s implementation with your check, then it would NOT catch the code that's in the wrong place and is being fixed by JDK-8288139. Setting the new _terminated field state exactly where I do so in this fix is what allowed me to prove that the code that I moved in JDK-8288139 is in the wrong place.

That's why I said except between these lines :) 

If you say this is the correct way to handle it for now, then it is.
Looks fine, thanks!

robehn (approved):

src/hotspot/share/runtime/thread.hpp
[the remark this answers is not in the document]
dcubed-ojdk: I do one more rename of the bug, the PR and the function name...

pchilano (approved):
Thanks Dan, looks good.

dcubed-ojdk:
@pchilano - Thanks for the re-review!

dcubed-ojdk:
The latest version's Mach5 Tier[1-7] looks good. Mach5 Tier8 was just started.

@dholmes-ora - please re-review when you get the chance.

src/hotspot/share/runtime/thread.inline.hpp
dholmes-ora: Sounds good. FTR this came in with the Thread-SMR changes which were reviewed over a period of many months with some "vigorous" discussion. :) I think by the time this particular code appeared the focus was on bigger issues.

src/hotspot/share/runtime/thread.inline.hpp
```diff
@@ -260,21 +260,24 @@ inline void JavaThread::set_done_attaching_via_jni() {
...
-  // Use load-acquire so that setting of _terminated by
-  // JavaThread::exit() is seen more quickly.
   TerminatedTypes l_terminated = Atomic::load_acquire(&_terminated);
-  return l_terminated == _thread_exiting || check_is_terminated(l_terminated);
+  return l_terminated == _thread_exiting ||
+         l_terminated == _thread_gc_barrier_detached ||
+         check_is_terminated(l_terminated);
+}
+
+inline bool JavaThread::is_oop_safe() const {
```
dholmes-ora, line 269: Do we want to document, and assert, that this is only intended for use by the current thread on itself? I can't imagine any scenario where we would need to ask whether an arbitrary thread is oop-safe. ??

dholmes-ora (approved):
Thanks for the updates Dan and your patience.

One minor query/suggestion below, but this is good as-is.

Thanks.

src/hotspot/share/runtime/thread.inline.hpp
dcubed-ojdk: @dholmes-ora - Thanks very much for the re-review.

I'll accept your "good as-is" and proceed with integration.

I think I've tweaked this fix enough at this point. Especially if we end up replacing
it with support in the GC subsystems instead.

PR #21, opened 2022-06-15
8288139: JavaThread touches oop after GC barrier is detached

dcubed-ojdk:
Update SharedRuntime::get_java_tid() to verify that the calling thread is safely
accessing its own threadObj(). This check uses the new is_gc_barrier_detached()
function added by [JDK-8288497](https://bugs.openjdk.org/browse/JDK-8288497) add support for JavaThread::is_gc_barrier_detached().

The above check was used to reproduce the failure mode without Shenandoah
and the remainder of the fix relocates the offending code from
ThreadsSMRSupport::remove_thread() to Threads::remove(). The work of
removed the 'tid' entry from the ThreadIdTable is still done under the
protection of the Threads_lock.

This fix along with the fix for JDK-8288497 has been tested in Mach5 Tier[1-8].
There are no related failures in Mach5 Tier[1-7]; Mach5 Tier8 is still running.

---------
### Progress

### Issue
 * [JDK-8288139](https://bugs.openjdk.org/browse/JDK-8288139): JavaThread touches oop after GC barrier is detached

### Reviewers
 * [Patricio Chilano Mateo](https://openjdk.org/census#pchilanomate) (@pchilano - **Reviewer**) ⚠️ Review applies to [a7302d19](https://git.openjdk.org/jdk19/pull/21/files/a7302d194bd7ca8a82367080e14837b429f4e59a)
 * [David Holmes](https://openjdk.org/census#dholmes) (@dholmes-ora - **Reviewer**) ⚠️ Review applies to [a7302d19](https://git.openjdk.org/jdk19/pull/21/files/a7302d194bd7ca8a82367080e14837b429f4e59a)
 * [Robbin Ehn](https://openjdk.org/census#rehn) (@robehn - **Reviewer**) ⚠️ Review applies to [1811bc14](https://git.openjdk.org/jdk19/pull/21/files/1811bc14a2003c564bc3c505a7d7fff59231db35)
 * [Erik Österlund](https://openjdk.org/census#eosterlund) (@fisk - **Reviewer**) ⚠️ Review applies to [1811bc14](https://git.openjdk.org/jdk19/pull/21/files/1811bc14a2003c564bc3c505a7d7fff59231db35)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/21/head:pull/21` \
`$ git checkout pull/21`

`$ git checkout pull/21` \
`$ git pull https://git.openjdk.org/jdk19 pull/21/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 21`

`$ git pr show -t 21`

</details>

<a href="https://git.openjdk.org/jdk19/pull/21.diff">https://git.openjdk.org/jdk19/pull/21.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -996,11 +996,12 @@ JRT_ENTRY_NO_ASYNC(void, SharedRuntime::register_finalizer(JavaThread* current, 
 JRT_END
 
 jlong SharedRuntime::get_java_tid(Thread* thread) {
-  if (thread != NULL) {
-    if (thread->is_Java_thread()) {
-      oop obj = JavaThread::cast(thread)->threadObj();
-      return (obj == NULL) ? 0 : java_lang_Thread::thread_id(obj);
-    }
+  if (thread != NULL && thread->is_Java_thread()) {
+    Thread* current = Thread::current();
+    guarantee(current != thread || !JavaThread::cast(thread)->is_gc_barrier_detached(),
+              "current cannot touch oops after its GC barrier is detached.");
+    oop obj = JavaThread::cast(thread)->threadObj();
+    return (obj == NULL) ? 0 : java_lang_Thread::thread_id(obj);
   }
   return 0;
 }
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -998,7 +998,7 @@ JRT_END
 jlong SharedRuntime::get_java_tid(Thread* thread) {
   if (thread != NULL && thread->is_Java_thread()) {
     Thread* current = Thread::current();
-    guarantee(current != thread || !JavaThread::cast(thread)->cannot_access_oops_safely(),
+    guarantee(current != thread || JavaThread::cast(thread)->is_oop_safe(),
               "current cannot touch oops after its GC barrier is detached.");
     oop obj = JavaThread::cast(thread)->threadObj();
     return (obj == NULL) ? 0 : java_lang_Thread::thread_id(obj);
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -998,7 +998,7 @@ JRT_END
 jlong SharedRuntime::get_java_tid(Thread* thread) {
   if (thread != NULL && thread->is_Java_thread()) {
     Thread* current = Thread::current();
-    guarantee(current != thread || !JavaThread::cast(thread)->is_gc_barrier_detached(),
+    guarantee(current != thread || !JavaThread::cast(thread)->cannot_access_oops_safely(),
               "current cannot touch oops after its GC barrier is detached.");
     oop obj = JavaThread::cast(thread)->threadObj();
     return (obj == NULL) ? 0 : java_lang_Thread::thread_id(obj);
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -3550,6 +3550,7 @@ jboolean Threads::is_supported_jni_version(jint version) {
   if (version == JNI_VERSION_1_8) return JNI_TRUE;
   if (version == JNI_VERSION_9) return JNI_TRUE;
   if (version == JNI_VERSION_10) return JNI_TRUE;
+  if (version == JNI_VERSION_19) return JNI_TRUE;
   return JNI_FALSE;
 }
 
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -3598,6 +3598,12 @@ void Threads::remove(JavaThread* p, bool is_daemon) {
     // StackWatermarkSet::on_safepoint(), which performs GC processing,
     // requiring the GC state to be alive.
     BarrierSet::barrier_set()->on_thread_detach(p);
+    if (p->is_exiting()) {
+      // If we got here via JavaThread::exit(), then we remember that the
+      // thread's GC barrier has been detached. We don't do this when we get
+      // here from another path, e.g., cleanup_failed_attach_current_thread().
+      p->set_terminated(JavaThread::_thread_gc_barrier_detached);
+    }
 
     assert(ThreadsSMRSupport::get_java_thread_list()->includes(p), "p must be present");
 
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -123,6 +123,7 @@
 #include "services/attachListener.hpp"
 #include "services/management.hpp"
 #include "services/memTracker.hpp"
+#include "services/threadIdTable.hpp"
 #include "services/threadService.hpp"
 #include "utilities/align.hpp"
 #include "utilities/copy.hpp"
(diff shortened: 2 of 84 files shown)


dcubed-ojdk:
@dholmes-ora, @Patricio and @robehn - Pinging you guys since we were all
involved with [JDK-8286830](https://bugs.openjdk.org/browse/JDK-8286830) ~HandshakeState should not touch oops

@fisk - Pinging you since you are interested in GC barriers.

pchilano (approved):
Looks good to me! I only have one comment but made it in 8288497 because it applies more to that change.

Thanks,
Patricio

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -996,11 +996,12 @@ JRT_ENTRY_NO_ASYNC(void, SharedRuntime::register_finalizer(JavaThread* current,
 JRT_END
 
 jlong SharedRuntime::get_java_tid(Thread* thread) {
```
dholmes-ora, line 998: Additional RFE: can we not declare this to take JavaThread and so avoid the checks and casts below?

src/hotspot/share/runtime/thread.cpp
```diff
@@ -3593,6 +3594,13 @@ void Threads::remove(JavaThread* p, bool is_daemon) {
   // that we do not remove thread without safepoint code notice
   { MonitorLocker ml(Threads_lock);
 
+    if (ThreadIdTable::is_initialized()) {
+      // This cleanup must be done before the current thread's GC barrier
+      // is detached since we need to touch the threadObj oop.
+      jlong tid = SharedRuntime::get_java_tid(p);
+      ThreadIdTable::remove_thread(tid);
+    }
+
```
dholmes-ora, line 3603: I don't like exposing `ThreadSMR` internal details like this. Can we at least make this a little `ThreadSMRSupport` method that we call from here.

Or ... can we just move `ThreadsSMRSupport::remove_thread(p);` to before `on_thread_detach`?

Or can we grab the pid before `on_thread_detach` and pass it to `ThreadsSMRSupport::remove_thread`?

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -996,11 +996,12 @@ JRT_ENTRY_NO_ASYNC(void, SharedRuntime::register_finalizer(JavaThread* current,
...
-  if (thread != NULL) {
-    if (thread->is_Java_thread()) {
-      oop obj = JavaThread::cast(thread)->threadObj();
-      return (obj == NULL) ? 0 : java_lang_Thread::thread_id(obj);
-    }
+  if (thread != NULL && thread->is_Java_thread()) {
+    Thread* current = Thread::current();
+    guarantee(current != thread || !JavaThread::cast(thread)->is_gc_barrier_detached(),
+              "current cannot touch oops after its GC barrier is detached.");
+    oop obj = JavaThread::cast(thread)->threadObj();
```
dholmes-ora, line 1003: I think the oop-touching-safety check should be done in `threadObj()` itself so that all callers are protected.

src/hotspot/share/runtime/thread.cpp
dcubed-ojdk: Ummmm... The code that I moved is not a TheadSMR internal detail. That code
is part of the M&M support that was added by Daniil in:

JDK-8185005 Improve performance of ThreadMXBean.getThreadInfo(long ids[], int maxDepth)
https://bugs.openjdk.org/browse/JDK-8185005

I don't remember why Daniil put that code in ThreadsSMRSupport::remove_thread(p),
but it is just as appropriate to put it in Threads::remove(). That M&M cleanup code
is part of the cleanup necessary when a JavaThread is removed. The important part
is to call that code after the Threads_lock is grabbed.

src/hotspot/share/runtime/sharedRuntime.cpp
dcubed-ojdk: I can research that and file a follow up RFE if necessary.

src/hotspot/share/runtime/sharedRuntime.cpp
dcubed-ojdk: The placement in SharedRuntime::get_java_tid() is because that is the failure
mode that this bug is about. My next stress testing experiment will be to put
a similar guarantee() in threadObj(), but I expect there to be more fan-out
from that placement.

This fix is tightly focued on the bug as it is reported because I plan to fix
it in JDK19.

dcubed-ojdk:
@dholmes-ora - Thanks for the review! I think I've replied to everything.
Please let me know if I missed something...

src/hotspot/share/runtime/thread.cpp
dholmes-ora: Thanks Dan, I had forgotten about that and just assumed it was ThreadSMR related. Moving it as you did is now not a concern.

src/hotspot/share/runtime/sharedRuntime.cpp
dholmes-ora: Yeah ... hmmm ... threadObj is still the better place for any safety check ... arguably it should be even deeper in OopHandle.

I'm not sure what fanout you are concerned about. If we find there are other invalid oop accesses then we would want to fix them in 19 - no?

dcubed-ojdk: I want to fix this bug in JDK19 since it is a blocker for @zhengyu123. I want to
keep this fix small and focused on the exact crash associated with the bug.

Yes, threadObj() or deeper is a better place. No argument, but I plan to do that
as a new hunt and I expect there to be many problems and I don't want to hold
up this fix for additional hunting. If I find more issues, then we'll determine how
to get them fixed in JDK19.

dcubed-ojdk:
The ThreadIdTable::remove_thread() call has to be done under the protection
of the Threads_lock if I remember correctly and the Threads_lock is not held
at that point in the code. There's some race condition with the other code
where I updated the comments.

dholmes-ora (approved):
Thanks Dan.

robehn:
Ok, seems fine, thanks.

robehn (approved):

fisk (approved):
Looks good.

dcubed-ojdk:
@dholmes-ora, @robehn and @fisk - Thanks for the re-review.
I'm doing another rename in JDK-8288497 so that will ripple into here...

pchilano (approved):
Still good.

dcubed-ojdk:
@pchilano - Thanks for the re-review!

dcubed-ojdk:
The latest version's Mach5 Tier[1-7] looks good. Mach5 Tier8 was just started.

dholmes-ora (approved):

dcubed-ojdk:
The Mach5 Tier8 run had no failures. All testing looks great.

dcubed-ojdk:
@dholmes-ora - Thanks for the re-review!

src/hotspot/share/runtime/sharedRuntime.cpp
dcubed-ojdk: New RFE:
[JDK-8289004](https://bugs.openjdk.org/browse/JDK-8289004) investigate if SharedRuntime::get_java_tid parameter should be a JavaThread*

src/hotspot/share/runtime/sharedRuntime.cpp
dcubed-ojdk: New bug:
[JDK-8289091](https://bugs.openjdk.org/browse/JDK-8289091) move oop safety check from SharedRuntime::get_java_tid() to JavaThread::threadObj()

PR #22, opened 2022-06-15
8287809: Revisit implementation of memory session

mcimadamore:
This is a JDK 19 clone of: https://github.com/openjdk/jdk/pull/9017

---------
### Progress

### Issue
 * [JDK-8287809](https://bugs.openjdk.org/browse/JDK-8287809): Revisit implementation of memory session

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/22/head:pull/22` \
`$ git checkout pull/22`

`$ git checkout pull/22` \
`$ git pull https://git.openjdk.org/jdk19 pull/22/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 22`

`$ git pr show -t 22`

</details>

<a href="https://git.openjdk.org/jdk19/pull/22.diff">https://git.openjdk.org/jdk19/pull/22.diff</a>

</details>


JornVernee (approved):
Already reviewed in the mainline repo.

mcimadamore:
I ended up simplifying this PR, by selecting some of the improvement from the previous patch:
* simplify handling of "readOnly" for memory segments
* Remove duplication between `sessionImpl` and `session` accessors, by making non-closeable views also be subclasses of `MemorySessionImpl`
* use better naming for the `checkValidState` methods - now the plain method is the method that should be used by all clients, while `checkValidStateRaw` is the method that should only be used in ScopedMemoryAccess
* unify exception handling, and fix issues around WrongThreadException not being handled correctly in ScopedMemoryAccess

src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -545,41 +524,36 @@ public static AbstractMemorySegmentImpl ofBuffer(Buffer bb) {
 
         AbstractMemorySegmentImpl bufferSegment = (AbstractMemorySegmentImpl)nioAccess.bufferSegment(bb);
         final MemorySessionImpl bufferSession;
-        int modes;
         if (bufferSegment != null) {
-            bufferSession = bufferSegment.sessionImpl();
-            modes = bufferSegment.mask;
+            bufferSession = bufferSegment.session;
         } else {
             bufferSession = MemorySessionImpl.heapSession(bb);
```
mcimadamore, line 530: I think this should be tweaked (as part of another patch). Buffer segments should have same lifecycle of array segments - and array segments have the global session. This is because an array segment always keeps the underlying array alive. I think we should enhance memory segment implementation (esp. that of native segments), so that it can be used to keep other objects alive. This would be useful also in the `loaderLookup` where we create another of those odd sessions - in that case we should just say that the segments returned by loader lookup have global scope, and keep the loader alive.

mcimadamore: I've filed a dependent PR for this:
https://github.com/openjdk/jdk19/pull/39

JornVernee:
Am I understanding correctly that, before calling `checkValidState` we also need to always call `baseSession()`? `checkValidState()` just calls `checkValidStateRaw` in a try/catch, and for non-closeable views this always throws.

There seem to be many cases where `baseSession()` is not called...

src/java.base/share/classes/java/lang/invoke/X-VarHandleSegmentView.java.template
```diff
@@ -128,18 +128,18 @@ final class VarHandleSegmentAs$Type$s extends VarHandleSegmentViewBase {
         VarHandleSegmentViewBase handle = (VarHandleSegmentViewBase)ob;
         AbstractMemorySegmentImpl bb = checkAddress(obb, base, handle.length, true);
 #if[floatingPoint]
-        $rawType$ rawValue = SCOPED_MEMORY_ACCESS.get$RawType$Unaligned(bb.sessionImpl(),
+        $rawType$ rawValue = SCOPED_MEMORY_ACCESS.get$RawType$Unaligned(bb.session(),
```
JornVernee, line 131: For instance, it's not clear to me why `baseSession()` is not called here.

JornVernee: It seems that, if `get$RawType$Unaligned` calls `checkValidStateRaw()` this will end up throwing an exception

src/java.base/share/classes/jdk/internal/foreign/abi/aarch64/macos/MacOsAArch64VaList.java
```diff
@@ -170,12 +165,12 @@ public MemoryAddress address() {
...
     public static non-sealed class Builder implements VaList.Builder {
 
-        private final MemorySession session;
+        private final MemorySessionImpl session;
         private final List<SimpleVaArg> args = new ArrayList<>();
 
         public Builder(MemorySession session) {
-            MemorySessionImpl.toSessionImpl(session).checkValidStateSlow();
-            this.session = session;
+            ((MemorySessionImpl)session).checkValidState();
```
JornVernee, line 172: Or here, if the memory session is a non-closeable view.

mcimadamore: I believe there was a wrong renaming with the IDE here, I will fix this

mcimadamore:
I've fixed the issues in the review comments, and further reduced the differences between the contents of this patch and mainline.

JornVernee (approved):

PR #24, opened 2022-06-15
8288534: Out of bound errors for memory segment access mentions wrong values

mcimadamore:
While playing with the API, I've realized that some of the out of bound error messgaes come out incorrectly.

This is because the bound check is performed as follows (to avoid overflow):

```
Objects.checkIndex(offset, this.length - length + 1);
```

So, if out-of-bounds access is detected, the resulting exception mentions the values of the first and second parameter, respectively - but since the second parameter is the result of a subtraction, it doesn't make sense.

The solution is not to use `Objects.checkIndex` directly, but, instead, drop down one level, and pass our own "IOOB formatter":

```
Preconditions.checkIndex(offset, this.length - length + 1, this);
```

Note that, in order to recover the correct values, the formatter needs to know the segment size - so I made `AbstractMemorySegment` act as a formatter (so we don't need to allocate in such a hot path).

The fix seems to bring back the desired messages - but I would like some confirmation that doing this won't disrupt intrinsification of the `checkIndex` method.

### DIFF
--- a/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/java.base/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -41,6 +41,7 @@ import java.nio.IntBuffer;
 import java.nio.LongBuffer;
 import java.nio.ShortBuffer;
 import java.util.*;
+import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.IntFunction;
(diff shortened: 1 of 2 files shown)


PaulSandoz:
I would expect passing an non-constant exception mapper should work. The C2 intrinsic never accesses the exception mapper argument, and a range check failure results in a deoptimzation back to the interpreter from which the exception mapper is then operated on. However, i think its worth just double checking generated code.

mcimadamore:
> I would expect passing an non-constant exception mapper should work. The C2 intrinsic never accesses the exception mapper argument, and a range check failure results in a deoptimzation back to the interpreter from which the exception mapper is then operated on. However, i think its worth just double checking generated code.

Both @iwanowww and @rwestrel have confirmed that that should be indeed the case. I did some test with PrintInlining with seemed to show the method as correctly inlined. For extra safety I will run all memory access benchmarks, and compare before/after.

mcimadamore:
Benchmarks looks all good except UnrolledAccess::handle_loop_instance. A quick PrintInlining revealed a missed inlining for `MemoryLayout.byteSize` which is called by `MemorySegment::getAtIndex`. This is an issue I spotted in the past, so we might just fix it now.

PR #28, opened 2022-06-16
8287982: Concurrent implicit attach from native threads crashes VM

AlanBateman:
If several native threads attach to the VM at around the same time, and before any threads get an automatically generated name are created, then the VM may crash attempting to access the thread status. The issue exists for native threads that attach explicitly with JNI AttachCurrentThread without a thread name, or native threads that attach implicitly by using a function pointer to do an up call.

The issue that raises its head periodically is that native threads that JNI attach do the initializaiton of the Thread object in the context of the attaching thread. Great care must be taken because Java code is executing in the context of a Thread that is not fully initialized. The right thing is probably to create the Thread object in another thread, using the service thread has been mentioned. The issue at this time arises when two or more native threads attempt to attach without thread names at around the same time. The first thread that needs an automatically generated name triggers the loading and initialization of a helper class.  If there are other threads attaching at the same time then they may have to wait on the monitor which can trigger the crash because the field holder with the thread status is not created at this time. Crashes in monitor enter and notify have been observed. Coleen has changed this code so that linking and initialization uses a mutex (JDK-8288064) so this specific crash doesn't duplicate in the main line. The short term fix for openjdk/jdk19 is to reorder the initialization so that field holder with the thread status is created before setting the name.

Creating a jtreg test with the conditions to duplicate this issue is complicated. The jtreg main wrapper creates the main thread with an automatically generated thread name before it runs the test main method. This is the reason that the test needs to launch a new VM with the right setup to exercise both explicit and implicit attach.

---------
### Progress

### Issue
 * [JDK-8287982](https://bugs.openjdk.org/browse/JDK-8287982): Concurrent implicit attach from native threads crashes VM

### Reviewers
 * [Robbin Ehn](https://openjdk.org/census#rehn) (@robehn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/28/head:pull/28` \
`$ git checkout pull/28`

`$ git checkout pull/28` \
`$ git pull https://git.openjdk.org/jdk19 pull/28/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 28`

`$ git pr show -t 28`

</details>

<a href="https://git.openjdk.org/jdk19/pull/28.diff">https://git.openjdk.org/jdk19/pull/28.diff</a>

</details>

### DIFF
--- a/make/test/JtregNativeJdk.gmk
+++ b/make/test/JtregNativeJdk.gmk
@@ -86,7 +86,7 @@ else
   BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libLinkerInvokerUnnamed := $(LIBCXX) -pthread
   BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libLinkerInvokerModule := $(LIBCXX) -pthread
   BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libLoaderLookupInvoker := $(LIBCXX) -pthread
-  BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libExplicitAttach := -pthread -ljvm
+  BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libExplicitAttach := -ljvm -pthread
   BUILD_JDK_JTREG_LIBRARIES_LDFLAGS_libImplicitAttach := -pthread
   BUILD_JDK_JTREG_EXCLUDE += exerevokeall.c
   ifeq ($(call isTargetOs, linux), true)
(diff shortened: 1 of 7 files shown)


dholmes-ora (approved):
Looks good.

And I confirmed that the VM code will correctly handle a null name from the current Java thread.

Thanks.

AlanBateman:
> And I confirmed that the VM code will correctly handle a null name from the current Java thread.

Thanks, I checked that one.

One thing that I'm wondering for the test is whether to move it to test/hotspot/jtreg/runtime/jni/ so it lives with the other tests for JNI. I had expected to find other tests for AttachCurrentThread in that location but we don't.

dholmes-ora:
> One thing that I'm wondering for the test is whether to move it to test/hotspot/jtreg/runtime/jni/ so it lives with the other tests for JNI. I had expected to find other tests for AttachCurrentThread in that location but we don't.

As this doesn't change hotspot code it seems more appropriate to have the regression test where it is. We have a number of tests that use AttachCurrentThread but very few actual regression tests for it.

Thanks.

PR #30, opened 2022-06-16
JDK-8288545: Missing space in error message

jonathan-gibbons:
Please review a minor update to fix the content of an error message.

As reported in JBS, the problem was partly that there was a missing space to separate two items, but also, the second item should not have been there anyway.

The fix here, for 19, is to suppress the second item.    Fixing the missing space will be part of a follow-up cleanup in mainline, that is too big to include here.

Note: there is no attempt in this fix to address whether the message should be generated at all.


hns (approved):
Looks good to me.

PR #31, opened 2022-06-16
8268398: 15% increase in JFR footprint in Noop-Base

egahlin:
Hi,

Could I have a review of a PR that skips bytecode instrumentation for container events and initialisation of Container.metrics() if the JVM isn't running in a container, or to be precise, if container support is not available. If a user specifies -XX:-UseContainerSupport, events are not emitted. This was the case before this PR, and maybe it should be fixed, but it's an enhancement and outside the scope of this bug.

Testing:
- jdk/jdk/jfr
- Manual testing in a container.

I tried to run test/hotspot/jtreg/containers/docker/TestJFREvents.java, but the test is broken (image can't be built, so test is skipped). 

This fix may not be sufficient to reduce footprint introduced by the container events, but [JDK-8282420: JFR: Remove event handlers](https://bugs.openjdk.org/browse/JDK-8282420) also reduced number of loaded classes which should reduce footprint in JDK 19. There are an ongoing enhancement work to generate bytecode and metadata at build time, which will help, but it will be integrated in JDK 20. 

Thanks
Erik

---------
### Progress

### Issue
 * [JDK-8268398](https://bugs.openjdk.org/browse/JDK-8268398): 15% increase in JFR footprint in Noop-Base

### Reviewers
 * [Jaroslav Bachorik](https://openjdk.org/census#jbachorik) (@jbachorik - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/31/head:pull/31` \
`$ git checkout pull/31`

`$ git checkout pull/31` \
`$ git pull https://git.openjdk.org/jdk19 pull/31/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 31`

`$ git pr show -t 31`

</details>

<a href="https://git.openjdk.org/jdk19/pull/31.diff">https://git.openjdk.org/jdk19/pull/31.diff</a>

</details>

### DIFF
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/JVM.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/JVM.java
@@ -620,4 +620,13 @@ public final class JVM {
      * @return the id, or a negative value if it does not exists.
      */
     public native long getTypeId(String name);
+
+    /**
+     * Returns {@code true}, if the JVM is running in a container, {@code false} otherwise.
+     * <p>
+     * If -XX:-UseContainerSupport has been specified, this method returns {@code false},
+     * which is questionable, but Container.metrics() returns {@code null}, so events
+     * can't be emitted anyway.
+     */
+    public native boolean isContainerized();
 }
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/MetadataRepository.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/MetadataRepository.java
@@ -218,6 +218,7 @@ public final class MetadataRepository {
         EventConfiguration configuration = newEventConfiguration(eventType, ec, settings);
         PlatformEventType pe = configuration.getPlatformEventType();
         pe.setRegistered(true);
+        // If class is instrumented or should not be instrumented, mark as instrumented.
         if (jvm.isInstrumented(eventClass) || !Utils.shouldInstrument(pe.isJDK(), pe.getName())) {
             pe.setInstrumented();
         }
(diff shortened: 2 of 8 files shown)


src/jdk.jfr/share/classes/jdk/jfr/internal/MetadataRepository.java
```diff
@@ -218,7 +218,7 @@ private EventConfiguration makeConfiguration(Class<? extends jdk.internal.event.
         EventConfiguration configuration = newEventConfiguration(eventType, ec, settings);
         PlatformEventType pe = configuration.getPlatformEventType();
         pe.setRegistered(true);
-        if (jvm.isInstrumented(eventClass)) {
+        if (jvm.isInstrumented(eventClass) || !Utils.shouldInstrument(pe.isJDK(), pe.getName())) {
```
jbachorik, line 221: This will call `setInstrumented()` even for events which are not really instrumented when they are not required to be instrumented, correct?

egahlin: Yes, I added comment to make it more clear.

jbachorik (approved):

src/jdk.jfr/share/classes/jdk/jfr/internal/JVM.java
```diff
@@ -620,4 +620,13 @@ public boolean hasNativeJFR() {
...
     public native long getTypeId(String name);
+
+    /**
+     * Returns {@code true}, if the JVM is running in a container, {@code false} otherwise.
+     * <p>
+     * If -XX:-UseContainerSupport has been specified, this method returns {@code false},
+     * which is questionable, but Container.metrics() returns {@code null}, so events
+     * can't be emitted anyway.
+     */
+    public native boolean isContainerized();
```
jerboaa, line 631: @egahlin I'm not sure this comment is correct, given https://bugs.openjdk.org/browse/JDK-8261242 and friends. It'll return true on almost all Linux systems (with cgroups pseudo filesystem available).

PR #39, opened 2022-06-17
8288697: Clarify lifecycle of buffer segments and loader lookup

mcimadamore:
This is a dependent PR containing a cleanup of the so called *heap sessions*. Heap sessions are used in two cases:

* when a buffer segment is created, to keep the original buffer instance reachable
* when a loader symbol lookup is created, to keep the classloader instance reachable

Spinning new sessions in these cases seems unnecessary, and inconsistent with what we do for segments backed by on-heap arrays, whose session is set to the global session. In that case, it's up to the segment to keep the underlying array reachable. I think that's a better model.

This patch adds the ability to attach Object references to native and mapped memory segments, so that we can attach loader/byte buffer instances to these segments, keeping them alive. This means that, in these cases we can go back to just use the global memory session, like we do for array segments.

This simplifies the implementation, makes the documentation more consistent, and also simplifies the user model, as it removes a concept (of heap session) that is not really documented anywhere. In fact, the javadoc for MemorySegment::ofBuffer claims, (wrongly!) that the session of the resulting segment is the global session - but that's not the case.

---------
### Progress

### Issue
 * [JDK-8288697](https://bugs.openjdk.org/browse/JDK-8288697): Clarify lifecycle of buffer segments and loader lookup

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/39/head:pull/39` \
`$ git checkout pull/39`

`$ git checkout pull/39` \
`$ git pull https://git.openjdk.org/jdk19 pull/39/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 39`

`$ git pr show -t 39`

</details>

<a href="https://git.openjdk.org/jdk19/pull/39.diff">https://git.openjdk.org/jdk19/pull/39.diff</a>

</details>


mcimadamore:
After thinking some more, I realized that going down this path might cause reachability issues when said segments are dereferenced (from ScopedMemoryAccess) or when they are passed to native function (using linker). Both system keep the session reachable, not the segment.

src/java.base/share/classes/java/lang/foreign/SymbolLookup.java
```diff
@@ -138,10 +139,12 @@ public interface SymbolLookup {
...
-     * returned by this method is backed by a {@linkplain MemorySession#asNonCloseable() non-closeable}, shared memory
-     * session which keeps the caller's class loader reachable. Therefore, libraries associated with the caller's class
-     * loader are kept loaded (and their symbols available) as long as a loader lookup for that class loader is reachable.
+     * <a href="../../../java/lang/ref/package.html#reachability">unreachable</a>. The returned loader lookup
+     * keeps the caller's class loader reachable. Therefore, libraries associated with the caller's class
+     * loader are kept loaded (and their symbols available) as long as a loader lookup for that class loader
+     * (or any of the symbols originated from it) is reachable.
+     * <p>
+     * The symbols obtained from the returned loader lookup are backed by the {@linkplain MemorySession#global() global session}.
      * <p>
```
AlanBateman, line 148: Does the spec update mean this should have a CSR?

mcimadamore: Yep - note that PR is in draft mode as we need to think through what changes we actually want to pursue.

PR #40, opened 2022-06-17
8288445: AArch64: C2 compilation fails with guarantee(!true || (true && (shift != 0))) failed: impossible encoding

dean-long:
The range for aarch64 vector right-shift is 1 to the element width.  This issue fixes the problem in the back-end.  There is a separate problem in the front-end that shift by 0 is not always optimized out.

---------
### Progress

### Issue
 * [JDK-8288445](https://bugs.openjdk.org/browse/JDK-8288445): AArch64: C2 compilation fails with guarantee(!true || (true && (shift != 0))) failed: impossible encoding

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**) ⚠️ Review applies to [0f927ce7](https://git.openjdk.org/jdk19/pull/40/files/0f927ce7b9e7d66d57de03ee71e2f18ee594195c)
 * [Hao Sun](https://openjdk.org/census#haosun) (@shqking - Author)
 * [Ningsheng Jian](https://openjdk.org/census#njian) (@nsjian - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/40/head:pull/40` \
`$ git checkout pull/40`

`$ git checkout pull/40` \
`$ git pull https://git.openjdk.org/jdk19 pull/40/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 40`

`$ git pr show -t 40`

</details>

<a href="https://git.openjdk.org/jdk19/pull/40.diff">https://git.openjdk.org/jdk19/pull/40.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
+++ b/test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
@@ -39,10 +39,10 @@ public class ShiftByZero {
 
     public static void bMeth() {
         int shift = i32[0];
+        // This loop is to confuse the optimizer, so that "shift" is
+        // optimized to 0 only after loop vectorization. 
         for (int i8 = 279; i8 > 1; --i8) {
             shift <<= 6;
-            >> This is to confuse the optimizer, so that "shift" is
-            >> optimized to 0 only after loop vectorization. 
         }
         // low 6 bits of shift are 0, so shift can be
         // simplified to constant 0
--- a/test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
+++ b/test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8288445
+ * @summary Test shift by 0
+ * @library /test/lib
+ * @run main compiler.codegen.ShiftByZero
+ * @run main/othervm -Xbatch -XX:-TieredCompilation
+ * -XX:+PrintOptoAssembly
+ * compiler.codegen.ShiftByZero
+ */
+
+package compiler.codegen;
+
+public class ShiftByZero {
+
+    public static final int N = 64;
+
+    public static int[] i32 = new int[N];
+
+    public static void bMeth() {
+        int shift = i32[0];
+        for (int i8 = 279; i8 > 1; --i8) {
+            shift <<= 6;
+        }
+        // low 6 bits of i1 are 0, so shift by i1 can be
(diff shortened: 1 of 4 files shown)


test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
```diff
@@ -0,0 +1,73 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/*
+ * @test
+ * @bug 8288445
+ * @summary Test shift by 0
+ * @library /test/lib
+ * @run main compiler.codegen.ShiftByZero
```
TobiHartmann, line 29: I don't think these two lines are needed.

test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
```diff
@@ -0,0 +1,73 @@
...
+                i32[i] >>= shift;
+            }
+            for (int i = 0; i < N; ++i) {
+                i32[i] <<= shift;
+            }
+        }
+    }
+
+    public static void main(String[] strArr) {
+        for (int i = 0; i < 20_000; i++ ) {
```
TobiHartmann, line 69:
```suggestion
        for (int i = 0; i < 20_000; i++) {
```

TobiHartmann (commented):
What instruction will the zero-shift be matched with then?

shqking:
Instead of introducing `immI_positive`, I wonder if we can generate `orr dst src` for zero shift count, just as the SVE part does. E.g., https://github.com/openjdk/jdk19/blob/master/src/hotspot/cpu/aarch64/aarch64_sve.ad#L3667

test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
dean-long: Thanks for catching it!

dean-long:
> What instruction will the zero-shift be matched with then?

The shift instruction that takes a vector register shift count rather than an immediate.

dean-long:
> Instead of introducing immI_positive, I wonder if we can generate orr dst src for zero shift count, just as the SVE part does

I considered using left-shift by 0, but I don't think it's worth it to optimize this case in the back-end.  If we really want to optimize shift by 0, I think it should be done in the front-end.

TobiHartmann (approved):
Looks good to me.

shqking:
> > Instead of introducing immI_positive, I wonder if we can generate orr dst src for zero shift count, just as the SVE part does
> 
> I considered using left-shift by 0, but I don't think it's worth it to optimize this case in the back-end. If we really want to optimize shift by 0, I think it should be done in the front-end.

Okay.

I think we may want to make the same update to the corresponding SVE part, e.g. https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/aarch64/aarch64_sve.ad#L3662, as it's better to align with the NEON and SVE implementations.

It may deserve one separate patch, not in this bugfix one.

dean-long:
Thanks Tobias.

dean-long:
@shqking Yes, that's a good idea for a separate RFE targeted for jdk20.

test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class ShiftByZero {
+
+    public static final int N = 64;
+
+    public static int[] i32 = new int[N];
+
+    public static void bMeth() {
+        int shift = i32[0];
+        for (int i8 = 279; i8 > 1; --i8) {
```
shqking, line 42: Why one loop is used here? I thought low 6 bits of shift can turn to be zeros in one single iteration. Is there anything I missed? Thanks.

dean-long: This is to confuse the optimizer, so that "shift" is optimized to 0 only after loop vectorization.  Otherwise, the back-end may never see such a value.

shqking: I see. Thanks.

shqking (approved):
LGTM (I'm not a Reviewer).

test/hotspot/jtreg/compiler/codegen/ShiftByZero.java
```diff
@@ -0,0 +1,73 @@
...
+public class ShiftByZero {
+
+    public static final int N = 64;
+
+    public static int[] i32 = new int[N];
+
+    public static void bMeth() {
+        int shift = i32[0];
+        // This loop is to confuse the optimizer, so that "shift" is
+        // optimized to 0 only after loop vectorization.
```
shqking, line 43: It seems that we should remove the trailing whitespace, as suggested by `jcheck`.
```suggestion
        // optimized to 0 only after loop vectorization.
```

shqking (approved):

nsjian (approved):
Looks good to me.

PR #42, opened 2022-06-18
8286580: serviceability/jvmti/vthread/GetSetLocalTest failed with assert: Not supported for heap frames

sspitsyn:
The JVMTI SetLocalXXX requires the target virtual thread suspended at a breakpoint or single step event.

This is the relevant statement from the "Local Variable" section:

`The SetLocalXXX functions may be used to set the value of a local variable in the topmost frame of a virtual thread suspended at a breakpoint or single step event.
`
This fix is to return the JVMTI_ERROR_OPAQUE_FRAME in cases when the target thread is not at a breakpoint or single step event. In this case the assert described in the bug report is avoided:

```
open/src/hotspot/share/runtime/vframe.cpp:300
# assert(stack_chunk() == __null) failed: Not supported for heap frames 
```
Also, this is an analysis from Ron:
```
The problem occurs because the thread is suspended at a safepoint poll on return in the oldest thawed frame. The safepoint happens after the frame is popped but before it returns to the return barrier, thawing the caller (and so the stack looks as if we're in enterSpecial). In that case the virtual thread's topmost frame is still frozen on the heap, even though it is mounted.

However, the spec says that a set local operation should succeed for the topmost frame of a mounted virtual thread only if the thread is suspended *at a breakpoint or a single-step event*, and I don't think we can stop at that safepoint in that case.

If so, the fix is simple: if we're trying to set, even if the virtual thread is mounted and the depth is zero, if the frame is a heap frame, we should return an OPAQUE_FRAME error. The test should be changed as well to accept such an error if the thread is suspended not at a breakpoint/single-step. 

```
Changes `GetSetLocalTest` test are taken from the repo-loom repository. It was an update from Leonid which is needed to reproduce this problem.

The fix was tested with thousands of runs of the `GetSetLocalTest` on Linux, Windows and MacOs.

Will also submit nsk.jvmti and nsk.jdi test runs on mach5.

---------
### Progress

### Issue
 * [JDK-8286580](https://bugs.openjdk.org/browse/JDK-8286580): serviceability/jvmti/vthread/GetSetLocalTest failed with assert: Not supported for heap frames

### Reviewers
 * [Leonid Mesnik](https://openjdk.org/census#lmesnik) (@lmesnik - **Reviewer**) ⚠️ Review applies to [108e3a46](https://git.openjdk.org/jdk19/pull/42/files/108e3a468f2b762e61a28f35fa66b1b92d2d836d)
 * [Ron Pressler](https://openjdk.org/census#rpressler) (@pron - Author) ⚠️ Review applies to [ca7c3b3a](https://git.openjdk.org/jdk19/pull/42/files/ca7c3b3a15be066e4f01e5ec6a9e8322fd2f6533)
 * [Chris Plummer](https://openjdk.org/census#cjplummer) (@plummercj - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/42/head:pull/42` \
`$ git checkout pull/42`

`$ git checkout pull/42` \
`$ git pull https://git.openjdk.org/jdk19 pull/42/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 42`

`$ git pr show -t 42`

</details>

<a href="https://git.openjdk.org/jdk19/pull/42.diff">https://git.openjdk.org/jdk19/pull/42.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -644,6 +644,13 @@ void VM_BaseGetOrSetLocal::doit() {
     return;
   }
   if (_set) {
+    if (fr.is_heap_frame()) { // we want this check after the check for JVMTI_ERROR_INVALID_SLOT
+      assert(_depth == 0 && Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
+      // the safepoint could be as we return to the return barrier but before we execute it (poll return)
+      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
+      return;
+    }
+
     // Force deoptimization of frame if compiled because it's
     // possible the compiler emitted some locals as constant values,
     // meaning they are not mutable.
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -645,7 +645,7 @@ void VM_BaseGetOrSetLocal::doit() {
   }
   if (_set) {
     if (fr.is_heap_frame()) { // we want this check after the check for JVMTI_ERROR_INVALID_SLOT
-      assert(_depth == 0 && Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
+      assert(Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
       // the safepoint could be as we return to the return barrier but before we execute it (poll return)
       _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
       return;
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -651,7 +651,6 @@ void VM_BaseGetOrSetLocal::doit() {
       // but the caller frame has not been thawed. We can't support a JVMTI SetLocal in the callee
       // frame at this point, because we aren't truly in the callee yet.
       // fr.is_heap_frame() is impossible if a continuation is at a single step or breakpoint.
-      // In such cases the top frames can't be frozen because of an agent callback frame.
       _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals are not fully supported in continuations
       return;
     }
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -611,7 +611,7 @@ void VM_BaseGetOrSetLocal::doit() {
 
   frame fr = _jvf->fr();
   if (_set && _depth != 0 && Continuation::is_frame_in_continuation(_jvf->thread(), fr)) {
-    _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
+    _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals are not fully supported in continuations
     return;
   }
 
@@ -646,8 +646,13 @@ void VM_BaseGetOrSetLocal::doit() {
   if (_set) {
     if (fr.is_heap_frame()) { // we want this check after the check for JVMTI_ERROR_INVALID_SLOT
       assert(Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
-      // the safepoint could be as we return to the return barrier but before we execute it (poll return)
-      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
+      // If the topmost frame is a heap frame, then it hasn't been thawed. This can happen
+      // if we are executing at a return barrier safepoint. The callee frame has been popped,
+      // but the caller frame has not been thawed. We can't support a JVMTI SetLocal in the callee
+      // frame at this point, because we aren't truly in the callee yet.
+      // fr.is_heap_frame() is impossible if a continuation is at a single step or breakpoint.
+      // In such cases the top frames can't be frozen because of an agent callback frame.
+      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals are not fully supported in continuations
       return;
     }
 
--- a/test/hotspot/jtreg/serviceability/jvmti/vthread/GetSetLocalTest/GetSetLocalTest.java
+++ b/test/hotspot/jtreg/serviceability/jvmti/vthread/GetSetLocalTest/GetSetLocalTest.java
@@ -100,7 +100,7 @@ public class GetSetLocalTest {
 
         GetSetLocalTest obj = new GetSetLocalTest();
 
-        for (int i = 0; i < 1000; i++) {
+        for (int i = 0; i < 200; i++) {
             obj.runTest();
         }
 
--- a/test/hotspot/jtreg/serviceability/jvmti/vthread/GetSetLocalTest/GetSetLocalTest.java
+++ b/test/hotspot/jtreg/serviceability/jvmti/vthread/GetSetLocalTest/GetSetLocalTest.java
@@ -35,8 +35,7 @@ import java.util.concurrent.*;
 public class GetSetLocalTest {
     private static final String agentLib = "GetSetLocalTest";
 
-    static final int MSG_COUNT = 600*1000;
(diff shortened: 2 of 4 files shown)


lmesnik (approved):

sspitsyn:
Thank you for review, Leonid!

src/hotspot/share/prims/jvmtiImpl.cpp
```diff
@@ -644,6 +644,13 @@ void VM_BaseGetOrSetLocal::doit() {
     return;
   }
   if (_set) {
+    if (fr.is_heap_frame()) { // we want this check after the check for JVMTI_ERROR_INVALID_SLOT
```
plummercj, line 647: It would be good to elaborate a bit more on this check. Say something about it being a "safepoint poll on return in the oldest thawed frame", and why that is a problem.

test/hotspot/jtreg/serviceability/jvmti/vthread/GetSetLocalTest/GetSetLocalTest.java
```diff
@@ -101,6 +99,10 @@ public static void main(String[] args) throws Exception {
         }
 
         GetSetLocalTest obj = new GetSetLocalTest();
-        obj.runTest();
+
+        for (int i = 0; i < 1000; i++) {
```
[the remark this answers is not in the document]
sspitsyn: No, it does not run quickly. However, this fix is needed to reproduce the issue.
In fact, it is to reproduce one more rare reproduceble issue:
8286836 Crash in serviceability/jvmti/vthread/GetSetLocalTest: No support for deferred values in continuations

src/hotspot/share/prims/jvmtiImpl.cpp
sspitsyn: Okay. I'd suggest this:
  if (_set) {
    if (fr.is_heap_frame()) { // we want this check after the check for JVMTI_ERROR_INVALID_SLOT
      assert(Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
      // The topmost frame of a mounted continuation can still be in the heap. This code is executed at safepoint.
      // The safepoint could be as we return to the return barrier but before we execute it (poll return).
      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
      return;
    }
Is it more clear?

sspitsyn: > I guess what I'm not understanding here is how/why "fr.is_heap_frame())" translates into "we are not single stepping or at a breakpoint". 
It is impossible to have `fr.is_heap_frame()` if we are at a single step or at a breakpoint. The frame has to be active or really executed at any event point.
I like your suggestion for the comment wording but I'm not sure if it is good enough.

> Also, I think the following comment is misleading:
I'll update this comment as well to make it more accurate.

sspitsyn: What about this update:

```
diff --git a/src/hotspot/share/prims/jvmtiImpl.cpp b/src/hotspot/share/prims/jvmtiImpl.cpp
index c34558c9ad5..8a7396e3cee 100644
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -611,7 +611,7 @@ void VM_BaseGetOrSetLocal::doit() {
 
   frame fr = _jvf->fr();
   if (_set && _depth != 0 && Continuation::is_frame_in_continuation(_jvf->thread(), fr)) {
-    _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
+    _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals are not fully unsupported in continuations
     return;
   }
 
@@ -646,8 +646,13 @@ void VM_BaseGetOrSetLocal::doit() {
   if (_set) {
     if (fr.is_heap_frame()) { // we want this check after the check for JVMTI_ERROR_INVALID_SLOT
       assert(Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
-      // the safepoint could be as we return to the return barrier but before we execute it (poll return)
-      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
+      // If the topmost frame is a heap frame, then it hasn't been thawed. This can happen
+      // if we are executing at a return barrier safepoint. The callee frame has been popped,
+      // but the caller frame has not been thawed. We can't support a JVMTI SetLocal in the callee
+      // frame at this point, because we aren't truly in the callee yet.
+      // fr.is_heap_frame() is impossible if a continuation is at a single step or breakpoint.
+      // In such cases the frames can't be frozen because a native callback frame is at the top.
+      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals are not fully unsupported in continuations
       return;
     }

```

sspitsyn: How about this change:
```
git diff src/hotspot/share/prims/jvmtiImpl.cpp
diff --git a/src/hotspot/share/prims/jvmtiImpl.cpp b/src/hotspot/share/prims/jvmtiImpl.cpp
index c34558c9ad5..8a7396e3cee 100644
--- a/src/hotspot/share/prims/jvmtiImpl.cpp
+++ b/src/hotspot/share/prims/jvmtiImpl.cpp
@@ -611,7 +611,7 @@ void VM_BaseGetOrSetLocal::doit() {
 
   frame fr = _jvf->fr();
   if (_set && _depth != 0 && Continuation::is_frame_in_continuation(_jvf->thread(), fr)) {
-    _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
+    _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals are not fully unsupported in continuations
     return;
   }
 
@@ -646,8 +646,13 @@ void VM_BaseGetOrSetLocal::doit() {
   if (_set) {
     if (fr.is_heap_frame()) { // we want this check after the check for JVMTI_ERROR_INVALID_SLOT
       assert(Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
-      // the safepoint could be as we return to the return barrier but before we execute it (poll return)
-      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals currently unsupported in continuations
+      // If the topmost frame is a heap frame, then it hasn't been thawed. This can happen
+      // if we are executing at a return barrier safepoint. The callee frame has been popped,
+      // but the caller frame has not been thawed. We can't support a JVMTI SetLocal in the callee
+      // frame at this point, because we aren't truly in the callee yet.
+      // fr.is_heap_frame() is impossible if a continuation is at a single step or breakpoint.
+      // In such cases the frames can't be frozen because a native callback frame is at the top.
+      _result = JVMTI_ERROR_OPAQUE_FRAME; // deferred locals are not fully unsupported in continuations
       return;
     }
```

test/hotspot/jtreg/serviceability/jvmti/vthread/GetSetLocalTest/GetSetLocalTest.java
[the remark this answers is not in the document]
sspitsyn: Decided to decrease the iterations number at L103 from 1000 to 200.
I hope, it should be a little bit quicker now.

pron (approved):

src/hotspot/share/prims/jvmtiImpl.cpp
```diff
@@ -644,6 +644,18 @@ void VM_BaseGetOrSetLocal::doit() {
...
+      assert(Continuation::is_frame_in_continuation(_jvf->thread(), fr), "sanity check");
+      // If the topmost frame is a heap frame, then it hasn't been thawed. This can happen
+      // if we are executing at a return barrier safepoint. The callee frame has been popped,
+      // but the caller frame has not been thawed. We can't support a JVMTI SetLocal in the callee
+      // frame at this point, because we aren't truly in the callee yet.
+      // fr.is_heap_frame() is impossible if a continuation is at a single step or breakpoint.
+      // In such cases the top frames can't be frozen because of an agent callback frame.
```
plummercj, line 654: "frames" should be "frame".  I don't understand the agent callback frame reference.

sspitsyn: Removed this sentence as unneeded.

plummercj (approved):

PR #49, opened 2022-06-20
8288759: GCC 12 fails to compile signature.cpp due to -Wstringop-overread

shipilev:
Trying to compile with GCC 12.1.1 (current Fedora Rawhide) yields this failure:

```
In file included from /home/test/shipilev-jdk/src/hotspot/share/utilities/globalDefinitions_gcc.hpp:35,
                 from /home/test/shipilev-jdk/src/hotspot/share/utilities/globalDefinitions.hpp:35,
                 from /home/test/shipilev-jdk/src/hotspot/share/memory/allocation.hpp:29,
                 from /home/test/shipilev-jdk/src/hotspot/share/classfile/classLoaderData.hpp:28,
                 from /home/test/shipilev-jdk/src/hotspot/share/precompiled/precompiled.hpp:34:
In function 'const void* memchr(const void*, int, size_t)',
    inlined from 'int SignatureStream::scan_type(BasicType)' at /home/test/shipilev-jdk/src/hotspot/share/runtime/signature.cpp:343:32,
    inlined from 'void SignatureStream::next()' at /home/test/shipilev-jdk/src/hotspot/share/runtime/signature.cpp:373:19,
    inlined from 'void SignatureIterator::do_parameters_on(T*) [with T = Fingerprinter]' at /home/test/shipilev-jdk/src/hotspot/share/runtime/signature.hpp:635:41,
    inlined from 'void SignatureIterator::do_parameters_on(T*) [with T = Fingerprinter]' at /home/test/shipilev-jdk/src/hotspot/share/runtime/signature.hpp:629:6,
    inlined from 'void Fingerprinter::compute_fingerprint_and_return_type(bool)' at /home/test/shipilev-jdk/src/hotspot/share/runtime/signature.cpp:169:19:
/usr/include/string.h:102:27: error: 'void* __builtin_memchr(const void*, int, long unsigned int)' specified bound [18446744073709486082, 0] exceeds maximum object size 9223372036854775807 [-Werror=stringop-overread]
  102 |   return __builtin_memchr (__s, __c, __n);
      |          ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

As @kimbarrett says: "The warning is indicating an actual problem with the code. The while loop on line 338 may terminate with end == limit if the string consists of just a sequence of '[' and then ends. If the loop ends for that reason, we later read base[limit], invoking UB as limit is the length of base."

Additional testing:
 - [x] Linux x86_64 fastdebug build with GCC 12.1.1
 - [x] Linux x86_64 fastdebug build with GCC 9.4.0

---------
### Progress

### Issue
 * [JDK-8288759](https://bugs.openjdk.org/browse/JDK-8288759): GCC 12 fails to compile signature.cpp due to -Wstringop-overread

### Reviewers
 * [Coleen Phillimore](https://openjdk.org/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to [e74cede8](https://git.openjdk.org/jdk19/pull/49/files/e74cede875913acffe11573c93f6df55996a3a8a)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/49/head:pull/49` \
`$ git checkout pull/49`

`$ git checkout pull/49` \
`$ git pull https://git.openjdk.org/jdk19 pull/49/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 49`

`$ git pr show -t 49`

</details>

<a href="https://git.openjdk.org/jdk19/pull/49.diff">https://git.openjdk.org/jdk19/pull/49.diff</a>

</details>


shipilev:
Any takers? :)

coleenp (approved):
Seems fine.

src/hotspot/share/utilities/compilerWarnings_gcc.hpp
```diff
@@ -59,6 +53,18 @@
 #define PRAGMA_NONNULL_IGNORED \
   PRAGMA_DISABLE_GCC_WARNING("-Wnonnull")
 
+// Disable -Wstringop-truncation which is introduced in GCC 8.
+// https://gcc.gnu.org/gcc-8/changes.html
+#if !defined(__clang_major__) && (__GNUC__ >= 8)
+#define PRAGMA_STRINGOP_TRUNCATION_IGNORED PRAGMA_DISABLE_GCC_WARNING("-Wstringop-truncation")
+#endif
```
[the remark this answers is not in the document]
kimbarrett: But note this part of the change shouldn't be done at all, per my comments elsewhere.

src/hotspot/share/runtime/signature.cpp
```diff
@@ -324,6 +324,8 @@ SignatureStream::~SignatureStream() {
   }
 }
 
+PRAGMA_DIAG_PUSH
+PRAGMA_STRINGOP_OVERREAD_IGNORED
```
kimbarrett, line 328: Don't make this change.  The warning is indicating an actual problem with the code.  The while loop on line 338 may terminate with `end == limit` if the string consists of just a sequence of '[' and then ends.  If the loop ends for that reason, we later read `base[limit]`, invoking UB as limit is the length of base.  As a proof of concept, adding
```
if (end >= limit) return limit;
```
after the while loop makes the warning go away.  I have no idea what the correct thing to do for this might be.  Returning limit might be wrong; I just used that to verify this issue is the source of the warning.

kimbarrett (changes requested):

src/hotspot/share/runtime/signature.cpp
shipilev: You're right, this is a legit warning. I see other code in `signature.cpp` that handles `JVM_SIGNATURE_ARRAY` scans and checks whether we ended up scanning the string completely. We should do the same here. Let me see...

shipilev:
@kimbarrett, @coleenp -- I redid the fix to fix the actual warning instead. I opted to return `limit` on the failure path + assert it does not actually happen in practice. It looks that returning `limit` is acceptable, as it rolls over to the "end of signature" on error. We can make that `fatal()` instead, but I don't like to penalize `release` builds unnecessarily.

kimbarrett (commented):
The proposed fix seems plausible, and addresses the warning.  I was a little worried that someone might come along later and try to change it to just
```
assert(end < limit, "invalid type");
```
but presumably that will get the warning again during testing.

I tried to think of a less contrived way to write this while still addressing the warning.  I haven't come up with anything better, assuming returning limit is okay.

But I'm entirely unfamiliar with the signature code, so don't know if returning limit is okay.  So don't count me as a reviewer for this change (and I won't hit the Approve button).

shipilev:
Closed in favor of https://github.com/openjdk/jdk/pull/9711

PR #54, opened 2022-06-21
8287379: Using @inheritDoc in an inapplicable context shouldn't crash javadoc

pavelrappo:
This rights the wrongs of JDK-8008768. For more information, see the respective CSR.

---------
### Progress
- [x] Change requires a CSR request to be approved

### Issues
 * [JDK-8287379](https://bugs.openjdk.org/browse/JDK-8287379): Using @inheritDoc in an inapplicable context shouldn't crash javadoc
 * [JDK-8288841](https://bugs.openjdk.org/browse/JDK-8288841): Using @inheritdoc in an inapplicable context shouldn't crash javadoc (**CSR**)

### Reviewers
 * [Jonathan Gibbons](https://openjdk.org/census#jjg) (@jonathan-gibbons - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/54/head:pull/54` \
`$ git checkout pull/54`

`$ git checkout pull/54` \
`$ git pull https://git.openjdk.org/jdk19 pull/54/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 54`

`$ git pr show -t 54`

</details>

<a href="https://git.openjdk.org/jdk19/pull/54.diff">https://git.openjdk.org/jdk19/pull/54.diff</a>

</details>

### DIFF
--- a/test/langtools/jdk/javadoc/doclet/testRelativeLinks/TestRelativeLinks.java
+++ b/test/langtools/jdk/javadoc/doclet/testRelativeLinks/TestRelativeLinks.java
@@ -60,6 +60,12 @@ public class TestRelativeLinks extends JavadocTester {
         // in the right places.
         checkOutput("pkg/C.html", true,
             """
+                <a href="relative-class-link.html">relative class link</a>""",
+            """
+                <a href="#class-fragment">fragment class link</a>""",
+            """
+                <a id="class-fragment">Class fragment</a>""",
+            """
                 <a href="relative-field-link.html">relative field link</a>""",
             """
                 <a href="relative-method-link.html">relative method link</a>""",
@@ -74,7 +80,11 @@ public class TestRelativeLinks extends JavadocTester {
             """
                 <a href="#package-fragment">package fragment link</a>""",
             """
-                <a id="package-fragment">Package fragment</a>""");
+                <a id="package-fragment">Package fragment</a>""",
+            """
+                <a href="relative-class-link.html">relative class link</a>""",
+            """
+                <a href="C.html#class-fragment">fragment class link</a>""");
 
         // subclass in same package
         checkOutput("pkg/D.html", true,
@@ -95,6 +105,10 @@ public class TestRelativeLinks extends JavadocTester {
 
         // INDEX PAGE
         checkOutput("index-all.html", true,
+            """
+                <a href="./pkg/relative-class-link.html">relative class link</a>""",
+            """
+                <a href="./pkg/C.html#class-fragment">fragment class link</a>""",
             """
                 <a href="./pkg/relative-field-link.html">relative field link</a>""",
             """
@@ -119,7 +133,11 @@ public class TestRelativeLinks extends JavadocTester {
             """
                 <a href="../pkg/relative-package-link.html">relative package link</a>""",
             """
-                <a href="../pkg/package-summary.html#package-fragment">package fragment link</a>""");
+                <a href="../pkg/package-summary.html#package-fragment">package fragment link</a>""",
+            """
+                <a href="../pkg/relative-class-link.html">relative class link</a>""",
+            """
+                <a href="../pkg/C.html#class-fragment">fragment class link</a>""");
 
         // CLASS_USE
         checkOutput("pkg/class-use/C.html", true,
--- a/test/langtools/jdk/javadoc/doclet/testRelativeLinks/pkg/C.java
+++ b/test/langtools/jdk/javadoc/doclet/testRelativeLinks/pkg/C.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -23,6 +23,13 @@
 
 package pkg;
 
+/**
+ * Here are two relative links in a class:
+ * <a href="relative-class-link.html">relative class link</a>,
+ * <a href="#class-fragment">fragment class link</a>.
+ *
+ * <a id="class-fragment">Class fragment</a>.
+ */
 public class C {
 
     /**


src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/Checker.java
```diff
@@ -263,6 +263,13 @@ public Void scan(DocCommentTree tree, TreePath p) {
         return null;
     }
 
+    // Checks if the passed tree path does NOT correspond to an entity, such as
+    // the overview file and doc-files/**/*.html files.
+    private boolean notPseudoElement(TreePath p) {
```
jonathan-gibbons, line 268: The use of a predicate named `not...` seems confusing and leads to various double negatives;  it would be easier to read if this was `isPseudoElement` even if that means an extra `!` at the call site.

test/langtools/jdk/javadoc/doclet/testRelativeLinks/TestRelativeLinks.java
```diff
@@ -59,12 +59,6 @@ public void testRelativeLinks() {
         // These relative paths should stay relative because they appear
         // in the right places.
         checkOutput("pkg/C.html", true,
-            """
-                <a href="relative-class-link.html">relative class link</a>""",
-            """
-                <a href="#class-fragment">fragment class link</a>""",
-            """
-                <a id="class-fragment">Class fragment</a>""",
```
jonathan-gibbons, line 67: Why are these lines (and similar, below) being deleted?

test/langtools/jdk/javadoc/doclet/testRelativeLinks/pkg/C.java
```diff
@@ -23,13 +23,6 @@
 
 package pkg;
 
-/**
- * Here are two relative links in a class:
- * <a href="relative-class-link.html">relative class link</a>,
- * <a href="#class-fragment">fragment class link</a>.
- *
- * <a id="class-fragment">Class fragment</a>.
- */
```
jonathan-gibbons, line 32: Why are these being deleted?

jonathan-gibbons (commented):
It worries me that we are removing support for a feature that had explicit positive tests.

test/langtools/jdk/javadoc/doclet/testRelativeLinks/pkg/C.java
pavelrappo: ditto

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/Checker.java
jonathan-gibbons: > It worries me that we are removing support for a feature that had explicit positive tests.

We discussed this offline at length.   The tests are not specifically for the feature, but instead, they are tests for a different feature (relative links) in a variety of situations, including the suspect case of `{@inheritDoc`.

It might be worth investigating whether we can write tests for relative links when `{@inheritDoc}` is used in a legal position, such as on a method declaration.

jonathan-gibbons:
This is a general response to the edits in `testRelativeLinks`.

There are various circumstances in which relative links need to be fixed up, all related to taking content containing links from one place and using it in another. These include:
* the first sentence of a class or interface comment, showing up in a package summary
* the first sentence of a package comment showing up in a module summary
* info being inherited from one file into another by means of `{@inherotDoc}`.
* etc ... info showing up on "other pages", like serialized-form, constant-values.

While I do not expect a full review of all these use cases, the _intent_ of the use of `{@inheritDoc}` in `testRelativeLinks` was good and well-meaning, even if the specific instances are now seen in hindsight as bad.  Bottom line: we should not simply remove the bad cases: we should replace them with good cases, meaning, test relative links with `{@inheritDoc}` used in method descriptions and/or tags.

pavelrappo:
> This is a general response to the edits in `testRelativeLinks`.
> 
> There are various circumstances in which relative links need to be fixed up, all related to taking content containing links from one place and using it in another. These include:
> 
> * the first sentence of a class or interface comment, showing up in a package summary,
> * the first sentence of a package comment showing up in a module summary,
> * info being inherited from one file into another by means of `{@inherotDoc}`,
> * etc ... info showing up on "other pages", like serialized-form, constant-values.
> 
> While I do not expect a full review of all these use cases, the _intent_ of the use of `{@inheritDoc}` in `testRelativeLinks` was good and well-meaning, even if the specific instances are now seen in hindsight as bad. Bottom line: we should not simply remove the bad cases: we should replace them with good cases, meaning, test relative links with `{@inheritDoc}` used in method descriptions and/or tags.

I accept that I nixed some tests overly aggressively. I carefully returned them back in 7d540c4. Note that since `{@inheritDoc}` is no longer allowed on a class or interface declaration, I cannot return all of the tests I deleted.

If fixing up relative links needs to be tested for a top-level declaration, then they have to be tested some other way, outside of the context of this PR; do you agree?

jonathan-gibbons:
> > This is a general response to the edits in `testRelativeLinks`.
> > There are various circumstances in which relative links need to be fixed up, all related to taking content containing links from one place and using it in another. These include:
> > 
> > * the first sentence of a class or interface comment, showing up in a package summary,
> > * the first sentence of a package comment showing up in a module summary,
> > * info being inherited from one file into another by means of `{@inherotDoc}`,
> > * etc ... info showing up on "other pages", like serialized-form, constant-values.
> > 
> > While I do not expect a full review of all these use cases, the _intent_ of the use of `{@inheritDoc}` in `testRelativeLinks` was good and well-meaning, even if the specific instances are now seen in hindsight as bad. Bottom line: we should not simply remove the bad cases: we should replace them with good cases, meaning, test relative links with `{@inheritDoc}` used in method descriptions and/or tags.
> 
> I accept that I nixed some tests overly aggressively. I carefully returned them back in [7d540c4](https://github.com/openjdk/jdk19/commit/7d540c4ede9e26b062f6fc977c5a310a95f1907d). Note that since `{@inheritDoc}` is no longer allowed on a class or interface declaration, I cannot return all of the tests I deleted.
> 
> If fixing up relative links needs to be tested for a top-level declaration, then they have to be tested some other way, outside of the context of this PR; do you agree?

That's not the right question.   The question is, how can we best replace the functionality of those tests. It's the relative link that's important, not the location.   Since the comment containing the relative link can no longer be placed on a top level declaration, they should be moved to a declaration where they are legal, such as on a method.

jonathan-gibbons:
> > That's not the right question. The question is, how can we best replace the functionality of those tests.

I looked at the complete test in more detail. The test is specifically about `{@inheritDoc}` in user tags, and the concern was about losing coverage.  But I see that we already have enough test cases for still-legal positions, so it is OK to remove the tests for now-illegal positions.

test/langtools/jdk/javadoc/doclet/InheritDocForUserTags/DocTest.java
```diff
@@ -64,16 +57,6 @@ public void test() {
         checkOutput(Output.STDERR, false, "at com.sun");
```
jonathan-gibbons, line 57: This is a general comment for a followup discussion and future work.

Should we have a standard default-on check in `JavadocTester` for "no stack traces"? That is, the equivalent of this line here. We could model such a check on `checkLinks` or `checkAccessibility` such that a hypothetical `checkNoCrashes` could be disabled in the (rare?) cases that they might be expected. (For example, checking the tool/doclet behavior when a crash does occur.)

jonathan-gibbons: This is a general comment for a followup discussion and future work.

Wh the bug and this test are about "no crashes", we should ensure that we have positive tests elsewhere for the actual output generated, assuming that no crash occurred.

jonathan-gibbons (approved):

test/langtools/jdk/javadoc/doclet/InheritDocForUserTags/DocTest.java
pavelrappo: We have such a check already, but it's used specifically for snippets. If you want, we could "hoist" that check from SnippetTester into JavadocTester. However, I would prefer doing that in a separate PR.

```
public class SnippetTester extends JavadocTester {
    ...

    /*
     * When checking for errors, it is important not to confuse one error with
     * another. This method checks that there are no crashes (which are also
     * errors) by checking for stack traces. We never expect crashes.
     */
    protected void checkNoCrashes() {
        checking("check crashes");
        Matcher matcher = Pattern.compile("\\s*at.*\\(.*\\.java:\\d+\\)")
                .matcher(getOutput(Output.STDERR));
        if (!matcher.find()) {
            passed("");
        } else {
            failed("Looks like a stacktrace: " + matcher.group());
        }
    }

    ...
```

PR #60, opened 2022-06-23
8247407: tools/jlink/plugins/CompressorPluginTest.java test failing

jaikiran:
Can I please get a review for this change which fixes an issue in `jdk.internal.jimage.ImageReader`? 

The `ImageReader` maintains a map of `nodes` which it uses to keep track of directories/resources from within the image. When a `findNode` is called, it leads to building the `Node` if the node isn't present in the map or if isn't complete (i.e. if the directory's immediate child nodes haven't been built). During this building of nodes, the code can end up creating a directory node and then visiting the child resources to create nodes for those resources. While doing so, the code currently doesn't check if a (child) node for the resource was already created previously (during a different call) and added to the parent directory's `children` list. As a result, it ends up adding the child node more than once.

One way to solve this would have been to make the `children` in `Directory` be a `Set`. I suspect it's currently a `List` so as to keep track of the order? If we did make it a `Set`, we could still keep the order using a `LinkedHashSet`. But changing this to a `Set` doesn't look feasible because there's a `List<Node> getChildren()` API on `Directory`, which would then have to create a copy of the `Set` to return as a `List`.

The commit in this PR instead prevents adding the child more than once by checking the `nodes` map for the presence of the child. This new check is added only for "resources" and not directories, because `makeDirectory` already skips a `Node` creation if the `nodes` has the corresponding entry (line 564 in the current PR).
Additionally an `assert` has been added to `addChild` of `Directory`.

A new jtreg test has been added which reproduces the issue without this change and verifies the fix.

This part of the code is very new to me, so if there's anything more to be considered, then please let me know.

tier1, tier2, tier3 testing is currently in progress.

---------
### Progress

### Issue
 * [JDK-8247407](https://bugs.openjdk.org/browse/JDK-8247407): tools/jlink/plugins/CompressorPluginTest.java test failing

### Reviewers
 * [Athijegannathan Sundararajan](https://openjdk.org/census#sundar) (@sundararajana - **Reviewer**) ⚠️ Review applies to [40bc7db8](https://git.openjdk.org/jdk19/pull/60/files/40bc7db8b2860cecfccad694e3bcc81b260f707c)
 * [Jim Laskey](https://openjdk.org/census#jlaskey) (@JimLaskey - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/60/head:pull/60` \
`$ git checkout pull/60`

`$ git checkout pull/60` \
`$ git pull https://git.openjdk.org/jdk19 pull/60/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 60`

`$ git pr show -t 60`

</details>

<a href="https://git.openjdk.org/jdk19/pull/60.diff">https://git.openjdk.org/jdk19/pull/60.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/jimage/ImageReader.java
+++ b/src/java.base/share/classes/jdk/internal/jimage/ImageReader.java
@@ -458,7 +458,11 @@ public final class ImageReader implements AutoCloseable {
                     makeDirectories(path);
                 } else { // a resource
                     makeDirectories(childloc.buildName(true, true, false));
-                    newResource(dir, childloc);
+                    // if we have already created a resource for this name previously, then don't
+                    // recreate it
+                    if (!nodes.containsKey(childloc.getFullName(true))) {
+                        newResource(dir, childloc);
+                    }
                 }
             });
             dir.setCompleted(true);
@@ -753,6 +757,7 @@ public final class ImageReader implements AutoCloseable {
         }
 
         void addChild(Node node) {
+            assert !children.contains(node) : "Child " + node + " already added";
             children.add(node);
         }
 
--- a/test/jdk/ProblemList.txt
+++ b/test/jdk/ProblemList.txt
@@ -695,7 +695,6 @@ javax/swing/JTable/8236907/LastVisibleRow.java 8284619 macosx-all
 
 # core_tools
 
-tools/jlink/plugins/CompressorPluginTest.java                   8247407 generic-all
 
 ############################################################################
 
--- a/test/jdk/tools/jimage/ImageReaderDuplicateChildNodesTest.java
+++ b/test/jdk/tools/jimage/ImageReaderDuplicateChildNodesTest.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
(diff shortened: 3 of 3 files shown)


sundararajana (approved):
LGTM. Nice catch!

AlanBateman:
Good sleuthing!

dcubed-ojdk:
This bug has gotten rather noisy in the last few days. We're up to 7 sightings
in < 24 hours. I'm going to ProblemList this test again in order to reduce the
noise in the JDK20 CI for this upcoming weekend.

dcubed-ojdk:
@jaikiran - The test has been ProblemListed again. Please merge jdk/jdk into this PR
and then redo your change to remove the test from the ProblemList. This PR is in kind
of a weird state because it currently contains a change to remove the test from the
ProblemList, but you later did that change separately for diagnostic purposes and did
not update this PR after that change was made. I'm reasonable certain that GitHub will
resolve the situation correctly when you merge jdk/jdk into this PR.

JimLaskey (approved):
LGTM

JimLaskey:
You'll now have to remove the test from the problem list.

jaikiran:
Thank you everyone for the reviews.

> You'll now have to remove the test from the problem list.

Hello Jim, this PR includes that removal. Once this PR is integrated and when jdk19 repo gets merged into jdk/jdk, that merge would take care of removing it from the JDK mainline code too.

PR #62, opened 2022-06-23
JDK-8288058: Broken links on constant-values page

jonathan-gibbons:
JDK 19:

Please review a medium-simple fix for some broken links on the constant-values page, from the "contents" index at the top of the page to the summary tables lower down on the page.

There are two root causes:

The packages are grouped into "groups" with headings based on abbreviated package names, meaning "at most two leading components of each package name". At one point there is a probable typo, when a full package name is used instead of the abbreviated name, thus causing the wrong "id" to be generated, thus causing broken links. The fix for this is obvious: use the correct name.

More seriously, the abbreviated package name is modeled using a PackageElement. While seemingly a good idea, the construction of these elements inherits the module element of the originating package. Then, the package elements are used to build a `Set` of headings, based on these package elements. The problem is that when the subpackages are found in different modules, they give rise to distinct "abbreviated package elements" in different modules and thus cause repeated instances of like-named headers in the output. This is easily visible as repeated headers for com.sun.* in (for example) the [JDK 18 API](https://docs.oracle.com/en/java/javase/18/docs/api/constant-values.html). The fix is to use a Set<String> to model the set of headers and associated links, instead of Set<PackageElement>.

Along with the two fixes described above, there is minor code cleanup, including the removal of some ill-considered code in both Workarounds and Utils. Apart from those classes, the changes are confined to the "constant summary" builder and writers. The use of a member called "first" and related shenanigans is a holdover from the days when the page was generated literally top-to-bottom with print statements. It is easily removed.

The only existing test for the constant values page was laughably insufficient, and was effectively just a regression test for a minor issue in JDK 1.4.2. The test is enhanced with a bunch of new test cases, that exercise different scenarios for the list of contents at the top of the page, and the links to the corresponding sections lower down. The new test cases leverage the implicit support for checking links in the generated output, although they do forcibly enable that support, just to make sure it is active.

When comparing the new JDK API docs against recent JDK API docs, the only changes are the expected changes in the links and headings on the constant-values.html page.

There is more cleanup that could be done to this page ... such as suppressing the "contents:" list when it is just a single entry ... but that is an enhancement that is out of scope for this bug fix.

---------
### Progress

### Issue
 * [JDK-8288058](https://bugs.openjdk.org/browse/JDK-8288058): Broken links on constant-values page

### Reviewers
 * [Pavel Rappo](https://openjdk.org/census#prappo) (@pavelrappo - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/62/head:pull/62` \
`$ git checkout pull/62`

`$ git checkout pull/62` \
`$ git pull https://git.openjdk.org/jdk19 pull/62/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 62`

`$ git pr show -t 62`

</details>

<a href="https://git.openjdk.org/jdk19/pull/62.diff">https://git.openjdk.org/jdk19/pull/62.diff</a>

</details>

### DIFF
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/ConstantsSummaryWriterImpl.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/ConstantsSummaryWriterImpl.java
@@ -26,7 +26,6 @@
 package jdk.javadoc.internal.doclets.formats.html;
 
 import java.util.Collection;
-import java.util.Set;
 
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.PackageElement;
@@ -97,20 +96,16 @@ public class ConstantsSummaryWriterImpl extends HtmlDocletWriter implements Cons
     }
 
     @Override
-    public void addLinkToPackageContent(PackageElement pkg,
-            Set<PackageElement> printedPackageHeaders, Content content) {
+    public void addLinkToPackageContent(String abbrevPackageName, Content content) {
         //add link to summary
         Content link;
-        if (pkg.isUnnamed()) {
+        if (abbrevPackageName.isEmpty()) {
             link = links.createLink(HtmlIds.UNNAMED_PACKAGE_ANCHOR,
                     contents.defaultPackageLabel, "");
         } else {
-            String parsedPackageName = utils.parsePackageName(pkg);
-            Content packageNameContent = Text.of(parsedPackageName + ".*");
-            link = links.createLink(DocLink.fragment(parsedPackageName),
+            Content packageNameContent = Text.of(abbrevPackageName + ".*");
+            link = links.createLink(DocLink.fragment(abbrevPackageName),
                     packageNameContent, "");
-            PackageElement abbrevPkg = configuration.workArounds.getAbbreviatedPackageElement(pkg);
-            printedPackageHeaders.add(abbrevPkg);
         }
         content.add(HtmlTree.LI(link));
     }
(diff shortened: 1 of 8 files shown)


src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/ConstantsSummaryWriterImpl.java
```diff
@@ -97,20 +96,16 @@ public Content getContentsHeader() {
...
         Content link;
-        if (pkg.isUnnamed()) {
+        if (abbrevPackageName.isEmpty()) {
             link = links.createLink(HtmlIds.UNNAMED_PACKAGE_ANCHOR,
                     contents.defaultPackageLabel, "");
         } else {
-            String parsedPackageName = utils.parsePackageName(pkg);
-            Content packageNameContent = Text.of(parsedPackageName + ".*");
-            link = links.createLink(DocLink.fragment(parsedPackageName),
+            Content packageNameContent = Text.of(abbrevPackageName + ".*");
```
pavelrappo, line 106: Separately. This is somewhat surprising to see that javadoc has been using `.*` to denote "subpackages" (or "related packages" as we call them elsewhere), whereas Java import statements use `.*` to denote immediately enclosed classes and interfaces, or static members thereof.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/ConstantsSummaryWriter.java
```diff
@@ -55,13 +55,10 @@ public interface ConstantsSummaryWriter {
...
-     * @param pkg                    the {@link PackageElement} to index.
-     * @param writtenPackageHeaders  the set of package headers that have already
-     *                               been indexed, we want to index utmost once.
-     * @param content                the content to which the link will be added
+     * @param abbrevPackageName the abbreviated package name
+     * @param content       the content to which the link will be added
      */
-    void addLinkToPackageContent(PackageElement pkg, Set<PackageElement> writtenPackageHeaders,
-                                 Content content);
+    void addLinkToPackageContent(String abbrevPackageName, Content content);
```
pavelrappo, line 61: It feels strange to downgrade a package representation from PackageElement to String. Especially considering that this interface expresses other program elements through javax.lang.model.

pavelrappo (commented):
I wonder why javadoc has been abbreviating packages down to two name components; Jon, do you have any ideas on that?

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/ConstantsSummaryWriter.java
jonathan-gibbons: Agreed it may seem strange.  Two points:
1. There is no representation for a package that ignores the enclosing module
2. These are not packages so much as headings, for the purpose of provides a "Contents" list at the head of the page, for ease of navigation.

test/langtools/jdk/javadoc/doclet/testConstantValuesPage/TestConstantValuesPage.java
```diff
@@ -41,14 +43,215 @@ public static void main(String... args) throws Exception {
...
+                    """,
+                """
+                    package p1.p2b.p3c;
+                    public class CC {
+                        public static final int ic = 1;
+                        public static final String sc = "string";
+                    }
+                    """,
+                """
+                    package p2;
```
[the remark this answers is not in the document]
jonathan-gibbons: yes, the constants page should only include the "included packages" given on the command line.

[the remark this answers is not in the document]
pavelrappo: I was asking because the test will still pass if that fourth (trailing) text block is deleted. So I wasn't sure what that text block was -- a part of the test or an artefact.

pavelrappo (approved):

PR #66, opened 2022-06-24
8288949: serviceability/jvmti/vthread/ContStackDepthTest/ContStackDepthTest.java failing

pron:
Please review the following bug fix:

`Continuation.enterSpecial` is a generated special nmethod (albeit not a Java method), with a well-known frame layout that calls `Continuation.enter`.

Because it is compiled, it resolves the call to `Continuation.enter` to its compiled version, if available. But this results in the compiled `Continuation.enter` being called even when the thread is in interp_only_mode.

This fix does three things:

1. When entering interp_only_mode, `Continuation::set_cont_fastpath_thread_state` will clear enterSpecial's resolved callsite to Continuation.enter.
2. In interp_only_mode, `SharedRuntime::resolve_static_call_C` will return `Continuation.enter`'s c2i entry rather than `verified_code_entry`.
3. In interp_only_mode, the c2i stub will not patch the callsite.

This fix isn't perfect, because a different thread, not in interp_only_mode, might patch the call. A longer-term solution is to create an "interpreted" version of `enterSpecial` and supporting an ad-hoc deoptimization. See https://bugs.openjdk.org/browse/JDK-8289128

Passes tiers 1-4 and Loom tiers 1-5.

### DIFF
--- a/src/hotspot/share/code/compiledIC.cpp
+++ b/src/hotspot/share/code/compiledIC.cpp
@@ -586,9 +586,6 @@ bool CompiledStaticCall::set_to_clean(bool in_use) {
   // Reset call site
   set_destination_mt_safe(resolve_call_stub());
 
-  // Do not reset stub here:  It is too expensive to call find_stub.
-  // Instead, rely on caller (nmethod::clear_inline_caches) to clear
-  // both the call and its stub.
   return true;
 }
 
--- a/src/hotspot/share/code/compiledMethod.cpp
+++ b/src/hotspot/share/code/compiledMethod.cpp
@@ -453,6 +453,19 @@ void CompiledMethod::clear_ic_callsites() {
   }
 }
 
+void CompiledMethod::clear_continuation_enter_special_inline_caches() {
+  assert (method()->is_continuation_enter_intrinsic(), "not Continuation.enterSpecial");
+
+  ResourceMark rm;
+  CompiledICLocker ml(this);
+  RelocIterator iter(this);
+  while(iter.next()) {
+    if (iter.type() == relocInfo::static_call_type) {
+      iter.reloc()->clear_inline_cache();
+    }
+  }
+}
+
 #ifdef ASSERT
 // Check class_loader is alive for this bit of metadata.
 class CheckClass : public MetadataClosure {
--- a/src/hotspot/share/runtime/continuation.cpp
+++ b/src/hotspot/share/runtime/continuation.cpp
@@ -311,6 +311,9 @@ void Continuation::set_cont_fastpath_thread_state(JavaThread* thread) {
   assert(thread != nullptr, "");
   bool fast = !thread->is_interp_only_mode();
   thread->set_cont_fastpath_thread_state(fast);
+  if (thread->is_interp_only_mode() && ContinuationEntry::enter_special() != nullptr) {
+    ContinuationEntry::enter_special()->clear_continuation_enter_special_inline_caches();
+  }
 }
 
 void Continuation::notify_deopt(JavaThread* thread, intptr_t* sp) {
--- a/src/hotspot/share/runtime/mutexLocker.cpp
+++ b/src/hotspot/share/runtime/mutexLocker.cpp
@@ -284,7 +284,7 @@ void mutex_init() {
   def(Heap_lock                    , PaddedMonitor, safepoint); // Doesn't safepoint check during termination.
   def(JfieldIdCreation_lock        , PaddedMutex  , safepoint);
 
-  def(CompiledIC_lock              , PaddedMutex  , nosafepoint);  // locks VtableStubs_lock, InlineCacheBuffer_lock
+  def(CompiledIC_lock              , PaddedMutex  , nosafepoint-1);  // locks VtableStubs_lock, InlineCacheBuffer_lock
   def(MethodCompileQueue_lock      , PaddedMonitor, safepoint);
   def(CompileStatistics_lock       , PaddedMutex  , safepoint);
   def(DirectivesStack_lock         , PaddedMutex  , nosafepoint);
--- a/src/hotspot/share/runtime/sharedRuntime.cpp
+++ b/src/hotspot/share/runtime/sharedRuntime.cpp
@@ -1560,10 +1560,28 @@ JRT_END
 // resolve a static call and patch code
 JRT_BLOCK_ENTRY(address, SharedRuntime::resolve_static_call_C(JavaThread* current ))
   methodHandle callee_method;
+  bool enter_special = false;
   JRT_BLOCK
     callee_method = SharedRuntime::resolve_helper(false, false, CHECK_NULL);
     current->set_vm_result_2(callee_method());
+
+    if (current->is_interp_only_mode()) {
+      RegisterMap reg_map(current, false);
+      frame stub_frame = current->last_frame();
+      assert(stub_frame.is_runtime_frame(), "must be a runtimeStub");
+      frame caller = stub_frame.sender(&reg_map);
+      enter_special = caller.cb() != NULL && caller.cb()->is_compiled()
+        && caller.cb()->as_compiled_method()->method()->is_continuation_enter_intrinsic();
+    }
   JRT_BLOCK_END
+
+  if (current->is_interp_only_mode() && enter_special) {
+    // enterSpecial is compiled and calls this method to resolve the call to Continuation::enter
+    // but in interp_only_mode we need to go to the interpreted entry
+    // The c2i won't patch in this mode -- see fixup_callers_callsite
+    return callee_method->get_c2i_entry();
+  }
+
   // return compiled code entry point after potential safepoints
   assert(callee_method->verified_code_entry() != NULL, " Jump to zero!");
   return callee_method->verified_code_entry();
@@ -1990,6 +2008,14 @@ JRT_LEAF(void, SharedRuntime::fixup_callers_callsite(Method* method, address cal
   // The check above makes sure this is a nmethod.
   CompiledMethod* nm = cb->as_compiled_method_or_null();
   assert(nm, "must be");
+
+  // Don't patch Continuation.enterSpecial if in interp_only mode.
+  // It's possible that another thread that isn't interp_only will patch enterSpecial,
+  // but we'll worry about that another time.
+  if (nm->method()->is_continuation_enter_intrinsic()) {
+    if (JavaThread::current()->is_interp_only_mode())
+      return;
+  }
 
   // Get the return PC for the passed caller PC.
   address return_pc = caller_pc + frame::pc_return_offset;
--- a/src/hotspot/share/code/compiledMethod.hpp
+++ b/src/hotspot/share/code/compiledMethod.hpp
@@ -374,6 +374,7 @@ public:
 
   virtual void clear_inline_caches();
   void clear_ic_callsites();
+  void clear_continuation_enter_special_inline_caches();
 
   // Execute nmethod barrier code, as if entering through nmethod call.
   void run_nmethod_entry_barrier();
--- a/src/hotspot/share/runtime/continuationEntry.cpp
+++ b/src/hotspot/share/runtime/continuationEntry.cpp
@@ -34,10 +34,12 @@
 
 int ContinuationEntry::_return_pc_offset = 0;
 address ContinuationEntry::_return_pc = nullptr;
+CompiledMethod* ContinuationEntry::_enter_special = nullptr;
 
 void ContinuationEntry::set_enter_code(CompiledMethod* cm) {
   assert(_return_pc_offset != 0, "");
   _return_pc = cm->code_begin() + _return_pc_offset;
+  _enter_special = cm;
 }
 
 ContinuationEntry* ContinuationEntry::from_frame(const frame& f) {
--- a/src/hotspot/share/runtime/continuationEntry.hpp
+++ b/src/hotspot/share/runtime/continuationEntry.hpp
@@ -52,7 +52,8 @@ public:
 
 public:
   static int _return_pc_offset; // friend gen_continuation_enter
-  static void set_enter_code(CompiledMethod* nm); // friend SharedRuntime::generate_native_wrapper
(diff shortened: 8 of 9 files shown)


src/hotspot/share/code/compiledIC.cpp
```diff
@@ -586,9 +586,6 @@ bool CompiledStaticCall::set_to_clean(bool in_use) {
   // Reset call site
   set_destination_mt_safe(resolve_call_stub());
 
-  // Do not reset stub here:  It is too expensive to call find_stub.
-  // Instead, rely on caller (nmethod::clear_inline_caches) to clear
-  // both the call and its stub.
```
pron, line 591: While at it, I noticed this comment, which appears to be out of date.

dean-long: I read that comment more as a warning, in case in the future someone wondered why we don't reset the stub here, and tried to add it.  So I would leave it.  I think it's still useful.

src/hotspot/share/code/compiledMethod.cpp
```diff
@@ -453,6 +453,19 @@ void CompiledMethod::clear_ic_callsites() {
...
 
+void CompiledMethod::clear_continuation_enter_special_inline_caches() {
+  assert (method()->is_continuation_enter_intrinsic(), "not Continuation.enterSpecial");
+
+  ResourceMark rm;
+  CompiledICLocker ml(this);
+  RelocIterator iter(this);
+  while(iter.next()) {
+    if (iter.type() == relocInfo::static_call_type) {
+      iter.reloc()->clear_inline_cache();
```
dean-long, line 464: This relies on code patching, and for correctness the change must be seen by the thread requesting interpreter-only mode.  If this was being done at a safepoint then it would probably be OK.  However, this code seems to be done using a handshake, so I'm not sure if the required serializing instruction is guaranteed to happen (see JDK-8220351).

src/hotspot/share/code/compiledIC.cpp
pron: ok, reverted.

src/hotspot/share/runtime/mutexLocker.cpp
```diff
@@ -284,7 +284,7 @@ void mutex_init() {
   def(Heap_lock                    , PaddedMonitor, safepoint); // Doesn't safepoint check during termination.
   def(JfieldIdCreation_lock        , PaddedMutex  , safepoint);
 
-  def(CompiledIC_lock              , PaddedMutex  , nosafepoint);  // locks VtableStubs_lock, InlineCacheBuffer_lock
+  def(CompiledIC_lock              , PaddedMutex  , nosafepoint-1);  // locks VtableStubs_lock, InlineCacheBuffer_lock
```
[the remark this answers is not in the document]
pron: The handshake lock, which is also nosafepoint.

dean-long:
The get_c2i_entry change seems safe enough, but the lock rank change and the code patching changes seem a little risky for jdk19.  I'm going to suggest some folks more familiar with handshakes, compiled ICs, and lock ranking to also look at this.

src/hotspot/share/code/compiledMethod.cpp
fisk: The change is not guaranteed to be seen on a concurrent thread, until the next global handshake operation completes.

pron: If that concurrent thread is in interp_only_mode, it also would have done the same patching. And if it isn't, then it's okay for it not to see this, but if it does see it, it will re-patch to compiled in c2i, as in the description.

src/hotspot/share/runtime/mutexLocker.cpp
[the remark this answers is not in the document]
coleenp: This should be ok, provided all the tests have been run.

It reduces the rank of other locks, but there's still room in the lock rank range (by 1), and there's an assert for that.
  // These locks have relative rankings, and inherit safepoint checking attributes from that rank.
  defl(InlineCacheBuffer_lock      , PaddedMutex  , CompiledIC_lock);
  defl(VtableStubs_lock            , PaddedMutex  , CompiledIC_lock);  // Also holds DumpTimeTable_lock
  defl(CodeCache_lock              , PaddedMonitor, VtableStubs_lock);
  defl(CompiledMethod_lock         , PaddedMutex  , CodeCache_lock);
  defl(CodeSweeper_lock            , PaddedMonitor, CompiledMethod_lock);

robehn:
Handshakes are per thread serialization points, so as Erik says, the thread going to interp mode will pick up the correct code, but other threads may or may not see it.
Lock rank change may be okay, to much code to trace just say yes for me.

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -1560,10 +1560,28 @@ JRT_END
 // resolve a static call and patch code
 JRT_BLOCK_ENTRY(address, SharedRuntime::resolve_static_call_C(JavaThread* current ))
   methodHandle callee_method;
+  bool enter_special = false;
```
sspitsyn, line 1563: One micro suggestion is to rename:  `enter_special => is_enter_special`.

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -1560,10 +1560,28 @@ JRT_END
...
+      enter_special = caller.cb() != NULL && caller.cb()->is_compiled()
+        && caller.cb()->as_compiled_method()->method()->is_continuation_enter_intrinsic();
+    }
   JRT_BLOCK_END
+
+  if (current->is_interp_only_mode() && enter_special) {
+    // enterSpecial is compiled and calls this method to resolve the call to Continuation::enter
+    // but in interp_only_mode we need to go to the interpreted entry
+    // The c2i won't patch in this mode -- see fixup_callers_callsite
+    return callee_method->get_c2i_entry();
```
sspitsyn, line 1582: Nit: Dots at the end of lines 1580-1581 would be nice to follow comments style in this file.

src/hotspot/share/runtime/sharedRuntime.cpp
```diff
@@ -1991,6 +2009,14 @@ JRT_LEAF(void, SharedRuntime::fixup_callers_callsite(Method* method, address cal
   CompiledMethod* nm = cb->as_compiled_method_or_null();
   assert(nm, "must be");
 
+  // Don't patch Continuation.enterSpecial if in interp_only mode.
+  // It's possible that another thread that isn't interp_only will patch enterSpecial,
+  // but we'll worry about that another time.
+  if (nm->method()->is_continuation_enter_intrinsic()) {
+    if (JavaThread::current()->is_interp_only_mode())
+      return;
+  }
```
sspitsyn, line 2018: Nit - micro simplification:
```
  if (nm->method()->is_continuation_enter_intrinsic() && JavaThread::current()->is_interp_only_mode()) {
      return;
  }
```

src/hotspot/share/runtime/continuation.cpp
```diff
@@ -311,6 +311,9 @@ void Continuation::set_cont_fastpath_thread_state(JavaThread* thread) {
   assert(thread != nullptr, "");
   bool fast = !thread->is_interp_only_mode();
   thread->set_cont_fastpath_thread_state(fast);
+  if (thread->is_interp_only_mode() && ContinuationEntry::enter_special() != nullptr) {
+    ContinuationEntry::enter_special()->clear_continuation_enter_special_inline_caches();
```
sspitsyn, line 315: Will this call impact all JavaThread's, not only the one passed in the argument?
Just want to understand this better.
Would it be worth to add a comment explaining this aspect (if applicable)?

sspitsyn:
How was this change tested?
In fact, it is not easy to estimate the total impact of this change. But it seems to be non-local.

pron:
Trying a new approach of having another entry into `enterSpecial`, used only when in interp-only-mode, and where the call to `Continuation.enter` always resolves to its interpreted version. This requires more platform-specific code, and also makes the frame appear not `frame::safe_for_sender` when at that callsite, but losing an async poll when in interp_only_mode doesn't seem to be a big issue, and the problem can be easily fixed as JFR is too eager to call `frame::safe_for_sender`.

robehn:
I think this is much better. I'll give it another round tomorrow.

src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
```diff
@@ -1015,30 +1015,72 @@ static void gen_continuation_enter(MacroAssembler* masm,
...
+    __ ldr(c_rarg1, Address(esp, Interpreter::stackElementSize*2));
+    __ ldr(c_rarg2, Address(esp, Interpreter::stackElementSize*1));
+    __ ldr(c_rarg3, Address(esp, Interpreter::stackElementSize*0));
+    __ push_cont_fastpath(rthread);
+
+    __ enter();
+    stack_slots = 2; // will be adjusted in setup
+    OopMap* map = continuation_enter_setup(masm, stack_slots);
+    // The frame is complete here, but we only record it for the compiled entry, so the frame would appear unsafe,
+    // but that's okay because at the very worst we'll miss an async sample, but we're in interp_only_mode anyeay.
```
dean-long, line 1050:
```suggestion
    // but that's okay because at the very worst we'll miss an async sample, but we're in interp_only_mode anyway.
```

src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
```diff
@@ -1171,19 +1213,23 @@ nmethod* SharedRuntime::generate_native_wrapper(MacroAssembler* masm,
...
     vmIntrinsics::ID iid = method->intrinsic_id();
     intptr_t start = (intptr_t)__ pc();
-    int vep_offset = ((intptr_t)__ pc()) - start;
+    int vep_offset = 0;
     int exception_offset = 0;
     int frame_complete = 0;
     int stack_slots = 0;
     OopMapSet* oop_maps =  new OopMapSet();
+    int interpreted_entry_offset = -1;
+    int compiled_entry_offset = -1;
```
dean-long, line 1222: `compiled_entry_offset` is unsed

src/hotspot/cpu/x86/sharedRuntime_x86_64.cpp
```diff
@@ -1471,17 +1526,21 @@ nmethod* SharedRuntime::generate_native_wrapper(MacroAssembler* masm,
...
     vmIntrinsics::ID iid = method->intrinsic_id();
     intptr_t start = (intptr_t)__ pc();
-    int vep_offset = ((intptr_t)__ pc()) - start;
+    int vep_offset = 0;
     int exception_offset = 0;
     int frame_complete = 0;
     int stack_slots = 0;
     OopMapSet* oop_maps =  new OopMapSet();
+    int interpreted_entry_offset = -1;
+    int compiled_entry_offset = -1;
```
dean-long, line 1535: `compiled_entry_offset` is unsed

src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
```diff
@@ -1015,30 +1015,72 @@ static void gen_continuation_enter(MacroAssembler* masm,
...
+    // The frame is complete here, but we only record it for the compiled entry, so the frame would appear unsafe,
+    // but that's okay because at the very worst we'll miss an async sample, but we're in interp_only_mode anyeay.
+
+    fill_continuation_entry(masm);
+
+    __ cmp(c_rarg2, (u1)0);
+    __ br(Assembler::NE, call_thaw);
+
+    address mark = __ pc();
+    __ trampoline_call1(resolve, NULL, false);
```
dean-long, line 1058: I don't think it's necessary to call the resolve stub when in interpreted mode.  Can't we just call the Method's c2i adapter just like the interpreter would?  I guess there might be a startup issue if the adapter hasn't been generated yet.

src/hotspot/cpu/x86/sharedRuntime_x86_64.cpp
```diff
@@ -1298,8 +1300,66 @@ static void gen_continuation_enter(MacroAssembler* masm,
...
+#ifdef ASSERT
+    Label is_interp_only;
+    __ cmpb(Address(r15_thread, JavaThread::interp_only_mode_offset()), 0);
+    __ jcc(Assembler::notEqual, is_interp_only);
+    __ stop("enterSpecial interpreter entry called when not in interp_only_mode");
+    __ bind(is_interp_only);
+#endif
+
+    __ pop(rax); // return address
+    // Read interpreter arguments into registers (this is an ad-hoc i2c adapter)
```
dean-long, line 1322: If I understand this correctly, this allows you to avoid creating an interpreted frame.  Pretty clever!

dean-long:
I like the new version.

src/hotspot/cpu/x86/sharedRuntime_x86_64.cpp
pron: Yeah, it's a hand-rolled i2c adapter. I thought of just calling `gen_i2c_adapter` in place, but that would have required changing it. If we had two separate nmethods, we could rely on the standard i2c, but having two nmethods for a single Method didn't seem safe at this time. We can revisit later.

src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
fisk: The callee belongs to the same class as the caller, so they should both be linked in this context.

fisk (approved):
Looks good.

robehn (approved):

src/hotspot/cpu/x86/sharedRuntime_x86_64.cpp
dean-long: I don't think the interpreter version needs to be an nmethod.  It could be more like the intrinsics generated by TemplateInterpreterGenerator::generate_method_entry, but let's revisit after jdk19.

PR #69, opened 2022-06-24
8289091: move oop safety check from SharedRuntime::get_java_tid() to JavaThread::threadObj()

dcubed-ojdk:
A trivial move of the oop safety check from SharedRuntime::get_java_tid() to
JavaThread::threadObj(). Also made adjustments to the threadObj() calls in
JavaThread::print_on_error() and JavaThread::get_thread_name_string() so
that we don't get secondary crashes when a JavaThread crashes after it has
detached the GC barrier.

Tested with Mach5 Tier[1-7]. A Mach5 Tier8 will be started this weekend.

---------
### Progress

### Issue
 * [JDK-8289091](https://bugs.openjdk.org/browse/JDK-8289091): move oop safety check from SharedRuntime::get_java_tid() to JavaThread::threadObj()

### Reviewers
 * [Robbin Ehn](https://openjdk.org/census#rehn) (@robehn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/69/head:pull/69` \
`$ git checkout pull/69`

`$ git checkout pull/69` \
`$ git pull https://git.openjdk.org/jdk19 pull/69/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 69`

`$ git pr show -t 69`

</details>

<a href="https://git.openjdk.org/jdk19/pull/69.diff">https://git.openjdk.org/jdk19/pull/69.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -2215,11 +2215,13 @@ const char* JavaThread::name() const  {
 // descriptive string if there is no set name.
 const char* JavaThread::get_thread_name_string(char* buf, int buflen) const {
   const char* name_str;
+#ifdef ASSERT
   Thread* current = Thread::current_or_null_safe();
   assert(current != nullptr, "cannot be called by a detached thread");
   if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
     // Only access threadObj() if current thread is not a JavaThread
     // or if it is a JavaThread that can safely access oops.
+#endif
     oop thread_obj = threadObj();
     if (thread_obj != NULL) {
       oop name = java_lang_Thread::name(thread_obj);
@@ -2237,6 +2239,7 @@ const char* JavaThread::get_thread_name_string(char* buf, int buflen) const {
     } else {
       name_str = Thread::name();
     }
+#ifdef ASSERT
   } else {
     // Current JavaThread has exited...
     if (current == this) {
@@ -2248,6 +2251,7 @@ const char* JavaThread::get_thread_name_string(char* buf, int buflen) const {
       name_str = Thread::name();
     }
   }
+#endif
   assert(name_str != NULL, "unexpected NULL thread name");
   return name_str;
 }
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -2147,11 +2147,10 @@ void JavaThread::print_name_on_error(outputStream* st, char *buf, int buflen) co
 // JavaThread::print() is that we can't grab lock or allocate memory.
 void JavaThread::print_on_error(outputStream* st, char *buf, int buflen) const {
   st->print("%s \"%s\"", type_name(), get_thread_name_string(buf, buflen));
-  Thread* current = Thread::current_or_null();
-  if (current != nullptr && (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe())) {
-    // Only access threadObj() if current thread is attached and
-    // if it is not a JavaThread or if it is a JavaThread that can safely
-    // access oops.
+  Thread* current = Thread::current();
+  if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
+    // Only access threadObj() if current thread is not a JavaThread
+    // or if it is a JavaThread that can safely access oops.
     oop thread_obj = threadObj();
     if (thread_obj != nullptr) {
       if (java_lang_Thread::is_daemon(thread_obj)) st->print(" daemon");
@@ -2214,12 +2213,8 @@ const char* JavaThread::name() const  {
 // descriptive string if there is no set name.
 const char* JavaThread::get_thread_name_string(char* buf, int buflen) const {
   const char* name_str;
-  Thread* current = Thread::current_or_null();
-  if (current == nullptr) {
-    // Current thread is not attached so it can't safely determine this
-    // JavaThread's name so use the default thread name.
-    name_str = Thread::name();
-  } else if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
+  Thread* current = Thread::current();
+  if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
     // Only access threadObj() if current thread is not a JavaThread
     // or if it is a JavaThread that can safely access oops.
     oop thread_obj = threadObj();
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -2147,9 +2147,13 @@ void JavaThread::print_name_on_error(outputStream* st, char *buf, int buflen) co
 // JavaThread::print() is that we can't grab lock or allocate memory.
 void JavaThread::print_on_error(outputStream* st, char *buf, int buflen) const {
   st->print("%s \"%s\"", type_name(), get_thread_name_string(buf, buflen));
-  if (is_oop_safe()) {
+  Thread* current = Thread::current_or_null();
+  if (current != nullptr && (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe())) {
+    // Only access threadObj() if current thread is attached and
+    // if it is not a JavaThread or if it is a JavaThread that can safely
+    // access oops.
     oop thread_obj = threadObj();
-    if (thread_obj != NULL) {
+    if (thread_obj != nullptr) {
       if (java_lang_Thread::is_daemon(thread_obj)) st->print(" daemon");
     }
   }
@@ -2210,7 +2214,14 @@ const char* JavaThread::name() const  {
 // descriptive string if there is no set name.
 const char* JavaThread::get_thread_name_string(char* buf, int buflen) const {
   const char* name_str;
-  if (is_oop_safe()) {
+  Thread* current = Thread::current_or_null();
+  if (current == nullptr) {
+    // Current thread is not attached so it can't safely determine this
+    // JavaThread's name so use the default thread name.
+    name_str = Thread::name();
+  } else if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
+    // Only access threadObj() if current thread is not a JavaThread
+    // or if it is a JavaThread that can safely access oops.
     oop thread_obj = threadObj();
     if (thread_obj != NULL) {
       oop name = java_lang_Thread::name(thread_obj);
@@ -2229,7 +2240,15 @@ const char* JavaThread::get_thread_name_string(char* buf, int buflen) const {
       name_str = Thread::name();
     }
   } else {
-    name_str = "<no-name - thread has exited>";
+    // Current JavaThread has exited...
+    if (current == this) {
+      // ... and is asking about itself:
+      name_str = "<no-name - current JavaThread has exited>";
+    } else {
+      // ... and it can't safely determine this JavaThread's name so
+      // use the default thread name.
+      name_str = Thread::name();
+    }
   }
   assert(name_str != NULL, "unexpected NULL thread name");
   return name_str;
--- a/src/hotspot/share/runtime/thread.cpp
+++ b/src/hotspot/share/runtime/thread.cpp
@@ -794,6 +794,9 @@ void JavaThread::set_threadOopHandles(oop p) {
 }
 
 oop JavaThread::threadObj() const {
+  Thread* current = Thread::current();
+  guarantee(current != this || JavaThread::cast(current)->is_oop_safe(),
+            "current cannot touch oops after its GC barrier is detached.");
   return _threadObj.resolve();
 }
 
@@ -2144,9 +2147,11 @@ void JavaThread::print_name_on_error(outputStream* st, char *buf, int buflen) co
 // JavaThread::print() is that we can't grab lock or allocate memory.
 void JavaThread::print_on_error(outputStream* st, char *buf, int buflen) const {
   st->print("%s \"%s\"", type_name(), get_thread_name_string(buf, buflen));
-  oop thread_obj = threadObj();
-  if (thread_obj != NULL) {
-    if (java_lang_Thread::is_daemon(thread_obj)) st->print(" daemon");
+  if (is_oop_safe()) {
+    oop thread_obj = threadObj();
+    if (thread_obj != NULL) {
+      if (java_lang_Thread::is_daemon(thread_obj)) st->print(" daemon");
+    }
   }
   st->print(" [");
   st->print("%s", _get_thread_state_name(_thread_state));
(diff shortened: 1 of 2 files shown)


dcubed-ojdk:
@dholmes-ora, @fisk, @pchilano, and @robehn - This is a followup from
[JDK-8288139](https://bugs.openjdk.org/browse/JDK-8288139) JavaThread touches oop after GC barrier is detached

robehn (approved):
Looks fine to me, thanks!

src/hotspot/share/runtime/thread.cpp
```diff
@@ -2144,9 +2147,11 @@ void JavaThread::print_name_on_error(outputStream* st, char *buf, int buflen) co
 // JavaThread::print() is that we can't grab lock or allocate memory.
 void JavaThread::print_on_error(outputStream* st, char *buf, int buflen) const {
   st->print("%s \"%s\"", type_name(), get_thread_name_string(buf, buflen));
-  oop thread_obj = threadObj();
-  if (thread_obj != NULL) {
-    if (java_lang_Thread::is_daemon(thread_obj)) st->print(" daemon");
+  if (is_oop_safe()) {
```
dholmes-ora, line 2150: This is wrong - `is_oop_safe` only has relevance when called on the current JavaThread. It is the current thread that must be oop_safe, not the target thread we are printing.

src/hotspot/share/runtime/thread.cpp
```diff
@@ -2205,22 +2210,26 @@ const char* JavaThread::name() const  {
 // descriptive string if there is no set name.
 const char* JavaThread::get_thread_name_string(char* buf, int buflen) const {
   const char* name_str;
-  oop thread_obj = threadObj();
-  if (thread_obj != NULL) {
-    oop name = java_lang_Thread::name(thread_obj);
-    if (name != NULL) {
-      if (buf == NULL) {
-        name_str = java_lang_String::as_utf8_string(name);
+  if (is_oop_safe()) {
```
dholmes-ora, line 2213: Same comment - it is the current thread that must be oop_safe. If the target thread has exited then we will detect that via the null threadObj.

dholmes-ora (changes requested):
Main change is fine but the "other adjustments" are not correct/appropriate. The state of the target thread is not the issue.

Thanks.

src/hotspot/share/runtime/thread.cpp
dcubed-ojdk: Agreed. I had that nagging feeling when I thought about this fix
while working on chores on Sunday...

src/hotspot/share/runtime/thread.cpp
dcubed-ojdk: Agreed.

dcubed-ojdk:
Also tested by temporarily reintroducing the bug fixed by:
[JDK-8288139](https://bugs.openjdk.org/browse/JDK-8288139)  JavaThread touches oop after GC barrier is detached
and verifying that the bad code is still detected and that the
hs_err_pid file looks good.

dholmes-ora (approved):
Hi Dan,
The additional checks to avoid secondary crashes seem like a lot of effort for little benefit, afterall the windows during which they would be applicable is quite small on the thread termination path. Somewhat ironically(?) the primary need for these additional checks would be when the current thread has already performed an unsafe oop access and so hit the guarantee and is now doing the thread dump for the hs_err file. Because of that I'm going to approve this, but in general I don't like us making the code jump through hoops for these kind of secondary failure avoidance issues. The current thread is the primary interest during a crash.

Thanks.

src/hotspot/share/runtime/thread.cpp
```diff
@@ -2144,9 +2147,15 @@ void JavaThread::print_name_on_error(outputStream* st, char *buf, int buflen) co
 // JavaThread::print() is that we can't grab lock or allocate memory.
 void JavaThread::print_on_error(outputStream* st, char *buf, int buflen) const {
   st->print("%s \"%s\"", type_name(), get_thread_name_string(buf, buflen));
-  oop thread_obj = threadObj();
-  if (thread_obj != NULL) {
-    if (java_lang_Thread::is_daemon(thread_obj)) st->print(" daemon");
+  Thread* current = Thread::current_or_null();
+  if (current != nullptr && (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe())) {
```
dholmes-ora, line 2151: Just realized, we can't have a null current thread if we are calling this as that is checked at a higher level. But we should be using `current_or_null_safe()` here as we could be in a signal-handling context.

src/hotspot/share/runtime/thread.cpp
```diff
@@ -2205,22 +2214,41 @@ const char* JavaThread::name() const  {
...
-  oop thread_obj = threadObj();
-  if (thread_obj != NULL) {
-    oop name = java_lang_Thread::name(thread_obj);
-    if (name != NULL) {
-      if (buf == NULL) {
-        name_str = java_lang_String::as_utf8_string(name);
+  Thread* current = Thread::current_or_null();
+  if (current == nullptr) {
+    // Current thread is not attached so it can't safely determine this
+    // JavaThread's name so use the default thread name.
```
dholmes-ora, line 2220: This should not be possible, but again `current_or_null_safe` should be used.

Though this code is used a lot in normal operation so the additional overhead of this is more significant than the print_on_error case.

I think this check should be moved to the caller if  needed (ie the print_on_error code), as in normal use it is not possible to fail this check.

dholmes-ora (changes requested):
Sorry Dan have to revoke my approval. There are still some issues here that need fixing and I don't like the impact on the code.

src/hotspot/share/runtime/thread.cpp
dcubed-ojdk: I used `Thread::current_or_null()` just to make the code more bullet proof. In all of
my testing I never ran into a case where `Thread::current()` returned `nullptr`.
I'm going to switch back to `Thread::current()` and remove the extra handling for
the `current == nullptr` case.

src/hotspot/share/runtime/thread.cpp
dcubed-ojdk: I'm going to switch back to `Thread::current()` and remove the extra handling for
the `current == nullptr` case.

dcubed-ojdk:
@dholmes-ora - Thanks for the re-review. I'm going to update the fix again and
switch back to using `Thread::current()` instead of `Thread::current_or_null()`.

The reason that this particular handling of a secondary failure is important is
that a secondary failure in printing the name of the failing thread will prevent
entries for following threads from being printed in the thread list in the hs_err_pid
file.

Here's an example to show placement:
```
Java Threads: ( => current thread )
  0x00007fd64d808a00 JavaThread "Unknown thread" [_thread_blocked, id=7427, stack(0x0000700003c18000,0x0000700003d18000)]
  0x00007fd64e80d400 JavaThread "Unknown thread" [_thread_blocked, id=43011, stack(0x0000700004433000,0x0000700004533000)]
  0x00007fd64e80e200 JavaThread "Unknown thread" [_thread_blocked, id=42755, stack(0x0000700004536000,0x0000700004636000)]
  0x00007fd64e80e800 JavaThread "Unknown thread" [_thread_blocked, id=42243, stack(0x0000700004639000,0x0000700004739000)]
  0x00007fd64e80ee00 JavaThread "Unknown thread" [_thread_blocked, id=22787, stack(0x000070000473c000,0x000070000483c000)]
  0x00007fd64e80f400 JavaThread "Unknown thread" [_thread_blocked, id=41731, stack(0x000070000483f000,0x000070000493f000)]
  0x00007fd64e81e800 JavaThread "Unknown thread" [_thread_blocked, id=41219, stack(0x0000700004942000,0x0000700004a42000)]
  0x00007fd64e81ee00 JavaThread "Unknown thread" [_thread_blocked, id=23299, stack(0x0000700004a45000,0x0000700004b45000)]
  0x00007fd650819000 JavaThread "Unknown thread" [_thread_blocked, id=40451, stack(0x0000700004b48000,0x0000700004c48000)]
  0x00007fd64e851400 JavaThread "Unknown thread" [_thread_blocked, id=23811, stack(0x0000700004c4b000,0x0000700004d4b000)]
  0x00007fd64d84e200 JavaThread "Unknown thread" [_thread_blocked, id=24067, stack(0x0000700004e51000,0x0000700004f51000)]
  0x00007fd64e81c600 JavaThread "Unknown thread" [_thread_blocked, id=39171, stack(0x0000700004f54000,0x0000700005054000)] _threads_hazard_ptr=0x00007fd64e0041b0
  0x00007fd64d80d000 JavaThread "Unknown thread" [_thread_blocked, id=38915, stack(0x0000700005057000,0x0000700005157000)]
=>0x00007fd650812000 JavaThread "<no-name - current JavaThread has exited>" [_thread_in_vm, id=24835, stack(0x000070000515a000,0x000070000525a000)]
```

The entry prefixed with "=>" is the crashing thread that is past the GC barrier
detach point. In this example, it happens to be the last thread in the `Java Threads:
( => current thread )` section of the hs_err_pid file. However, if the failing thread
happened to be earlier in the list and we didn't prevent the secondary error, then
we would be missing entries from the section.

dcubed-ojdk:
Testing the V02 patch with Mach5 Tier1 now.

Also tested by temporarily reintroducing the bug fixed by:
[JDK-8288139](https://bugs.openjdk.org/browse/JDK-8288139) JavaThread touches oop after GC barrier is detached
and verifying that the bad code is still detected and that the
hs_err_pid file looks good.

dcubed-ojdk:
@dholmes-ora - please re-review when you get the chance.
Mach5 Tier[1-4] has passed and Tier[5-7] are running along nicely.

src/hotspot/share/runtime/thread.cpp
dholmes-ora: If the code can be invoked in the context of a signal handler, as can happen for the error reporting path, then you must use `Thread::current_or_null_safe()` or risk introducing a deadlock or secondary crash.
It occurs to me that this may mean the existing safety-check is not correct because it may end up being checked in that context too.

dholmes-ora:
> The reason that this particular handling of a secondary failure is important is
that a secondary failure in printing the name of the failing thread will prevent
entries for following threads from being printed in the thread list in the hs_err_pid
file.

Yes I realise that a secondary crash loses some information, but my contention is that:

a) the likelihood of crashing after detaching the GC barrier is very, very small; and
b) in such a crash it is only the crashing thread that is really of interest, not the other threads in the system

So to me trying to make secondary crash handling more robust in the current case is not worth the cost of the extra checks. If the checks were only in the crash reporting path then that would be okay, but not when they impact normal code execution.  Sorry.

src/hotspot/share/runtime/thread.cpp
```diff
@@ -2147,11 +2147,10 @@ void JavaThread::print_name_on_error(outputStream* st, char *buf, int buflen) co
...
   st->print("%s \"%s\"", type_name(), get_thread_name_string(buf, buflen));
-  Thread* current = Thread::current_or_null();
-  if (current != nullptr && (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe())) {
-    // Only access threadObj() if current thread is attached and
-    // if it is not a JavaThread or if it is a JavaThread that can safely
-    // access oops.
+  Thread* current = Thread::current();
+  if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
+    // Only access threadObj() if current thread is not a JavaThread
+    // or if it is a JavaThread that can safely access oops.
```
dholmes-ora, line 2153: How can a non-JavaThread safely access the oop? Is the only safe case the VMThread at a safepoint?

src/hotspot/share/runtime/thread.cpp
```diff
@@ -2144,9 +2147,14 @@ void JavaThread::print_name_on_error(outputStream* st, char *buf, int buflen) co
 // JavaThread::print() is that we can't grab lock or allocate memory.
 void JavaThread::print_on_error(outputStream* st, char *buf, int buflen) const {
   st->print("%s \"%s\"", type_name(), get_thread_name_string(buf, buflen));
-  oop thread_obj = threadObj();
-  if (thread_obj != NULL) {
-    if (java_lang_Thread::is_daemon(thread_obj)) st->print(" daemon");
+  Thread* current = Thread::current();
+  if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
```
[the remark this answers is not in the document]
dcubed-ojdk: The purpose of the guarantee() in threadObj() is catch bad calls to
threadObj() that a thread makes after it has passed the GC barrier detach
point. Returning nullptr from threadObj() would defeat this purpose.

src/hotspot/share/runtime/thread.cpp
dcubed-ojdk: The purpose of the `if (!current->is_Java_thread() ||` part of the if-statement
is to allow the code to work as it did before for the non-JavaThread case. Before
this fix, if a non-JavaThread called into this code, then it was allowed to execute
this code. I've preserved that behavior and I've see no failures that indicate that
this is a problem.

Do I know what non-JavaThreads might wander in here? No I don't.

dcubed-ojdk:
@dholmes-ora - I'm very glad that I resisted putting the check in `threadObj()`
when I made the changes for:

[JDK-8288139](https://bugs.openjdk.org/browse/JDK-8288139) JavaThread touches oop after GC barrier is detached

The review for that PR took too long and clearly this discussion would have made
it take longer. JDK-8288139 is fixed and Zhengyu can make progress on his issue.

I don't see a good way to make forward progress in this PR. It doesn't look like you
and I will reach a solution that's acceptable to both of us.

The only good news is that my many rounds of testing on this fix have convinced me
threadObj() is not being abused in the current code base except for the very narrow
case where we crash (for whatever reason) after the GC barrier is detached. Should
the GC team chose to put in an appropriate trap in their code that catches a thread
accessing an oop after the GC barrier is detached, then threadObj() will be automatically
sanity checked by that code.

Thanks for your time in doing several rounds of review.

src/hotspot/share/runtime/thread.cpp
dholmes-ora: I realise this is pre-existing behaviour. It seems odd that a non-JavaThread can touch an oop when an exiting JavaThread cannot - how are they different? Or do we just cross our fingers and hope for the best with a non-JavaThread because it is rare? Perhaps @fisk can explain?

dholmes-ora:
> I'm very glad that I resisted putting the check in threadObj() when I made the changes for ...

The problem has not been (until now) what you put in threadObj but the other checks you decided to put in.

But now the issue has been raised, any code that can be executed in the context of a signal handler must use `Thread::current_or_null_safe()`. I don't know if the sharedRuntime code falls into that category but a call to `threadObj()` certainly can.

Sorry this has not been smooth sailing.

dcubed-ojdk:
I've re-read the history behind:

[JDK-8132510](https://bugs.openjdk.org/browse/JDK-8132510) Replace ThreadLocalStorage with compiler/language-based thread-local variables

which is the fix that introduced `Thread::current_or_null_safe()`. Wow does that fix
and the code review process bring back memories. I remember the struggle to get
the fix in before JDK9 FC... 10 releases ago... yikes!

I'm mulling and researching on what to do...

dcubed-ojdk:
This latest version has only been tested with Mach5 Tier1 so far.

dcubed-ojdk:
In JavaThread::get_thread_name_string() and when you look at the PR with
white space changes disabled, you'll see this new code which is executed
in the "normal code" case:

```
  Thread* current = Thread::current_or_null_safe();
  assert(current != nullptr, "cannot be called by a detached thread");
  if (!current->is_Java_thread() || JavaThread::cast(current)->is_oop_safe()) {
    // Only access threadObj() if current thread is not a JavaThread
    // or if it is a JavaThread that can safely access oops.
```
so I've switched to `Thread::current_or_null_safe` and added an
assert() for the `current` value not being `nullptr`.

dcubed-ojdk:
@dholmes-ora - please re-review when you get the chance. I've started a
new round of testing in Mach5.

robehn (approved):
Still good.

dholmes-ora (approved):
Thanks for the changes Dan. I can live with this version. :)

Thanks.

dcubed-ojdk:
@robehn and @dholmes-ora - Thanks for the re-reviews.

/integrate

PR #70, opened 2022-06-24
JDK-8275784: Bogus warning generated for record with compact constructor

jonathan-gibbons:
Please review a simple change to disable the check for missing `@param` tags in the case of a compact constructor for a record. In this case, there is no explicit argument list: it is inferred from the  list of record components, and so there is no need to check for `@param` tags.

---------
### Progress

### Issue
 * [JDK-8275784](https://bugs.openjdk.org/browse/JDK-8275784): Bogus warning generated for record with compact constructor

### Reviewers
 * [Pavel Rappo](https://openjdk.org/census#prappo) (@pavelrappo - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/70/head:pull/70` \
`$ git checkout pull/70`

`$ git checkout pull/70` \
`$ git pull https://git.openjdk.org/jdk19 pull/70/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 70`

`$ git pr show -t 70`

</details>

<a href="https://git.openjdk.org/jdk19/pull/70.diff">https://git.openjdk.org/jdk19/pull/70.diff</a>

</details>

### DIFF
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/Checker.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/Checker.java
@@ -267,7 +267,7 @@ public class Checker extends DocTreePathScanner<Void, Void> {
         return null;
     }
 
-    private  boolean isCanonicalRecordConstructor(ExecutableElement ee) {
+    private boolean isCanonicalRecordConstructor(ExecutableElement ee) {
         TypeElement te = (TypeElement) ee.getEnclosingElement();
         if (te.getKind() != ElementKind.RECORD) {
             return false;
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/Checker.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/Checker.java
@@ -33,6 +33,7 @@ import java.util.Deque;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclint/Checker.java
```diff
@@ -263,6 +267,31 @@ public Void scan(DocCommentTree tree, TreePath p) {
         return null;
     }
 
+    private  boolean isCanonicalRecordConstructor(ExecutableElement ee) {
```
pavelrappo, line 270: Remove extra whitespace between `private` and `boolean`.

I would expect javax.lang.model to provide this functionality. Unfortunately, I couldn't find it. On the other hand, a very similar method exists in jdk.javadoc.internal.doclets.toolkit.util.Utils. It would be nice if we could avoid any duplication.

jonathan-gibbons: I don't see how we can avoid duplication at this time.

jonathan-gibbons: It's not actually clear that `java.lang.model` has the info to provide this method, since the compact constructor is a syntactic form, that has disappeared by the time we get to the `ExecutableElement` for the constructor.  Sure, we could provide the method, but it's not clear we can detect the difference between the compact form and the use of explicit parameters.  I guess we could maybe put it on `Trees`.

jonathan-gibbons: [JDK-8289249](https://bugs.openjdk.org/browse/JDK-8289249)

pavelrappo (approved):

PR #72, opened 2022-06-25
8289069: Very slow C1 arraycopy jcstress tests after JDK-8279886Size bitmaps appropriately

veresov:
I used BlockBegin::number_of_blocks() to size the bitmaps, however that is a total number of blocks. Since mark_loops() is called after every inlining (for every inlinee - no need to reanalyze the whole method), the bitmaps get progressively larger, and have to be zeroed. That makes the complexity quadratic. The solution is to appropriately size the bitmaps and keep the whole thing linear. Tests look good.

### DIFF
--- a/src/hotspot/share/c1/c1_GraphBuilder.cpp
+++ b/src/hotspot/share/c1/c1_GraphBuilder.cpp
@@ -66,7 +66,9 @@ class BlockListBuilder {
   GrowableArray<ResourceBitMap> _loop_map; // caches the information if a block is contained in a loop
   int            _next_loop_index;     // next free loop number
   int            _next_block_number;   // for reverse postorder numbering of blocks
+  int            _block_id_start;
 
+  int           bit_number(int block_id) const   { return block_id - _block_id_start; }
   // accessors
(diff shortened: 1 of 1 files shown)


src/hotspot/share/c1/c1_GraphBuilder.cpp
```diff
@@ -406,7 +410,9 @@ void BlockListBuilder::mark_loops() {
   // a node with some loop bit set that would then propagate back and be never cleared because we would
   // never go back through the original loop header. Therefore if there are any irreducible loops the bits in the states
   // for these loops are going to propagate back to the root.
-  BitMap& loop_state = mark_loops(_bci2block->at(0), false);
+  BlockBegin* start = _bci2block->at(0);
```
TobiHartmann, line 413: There is a typo in line 408, that you might want to fix as well `irriducible` -> `irreducible`.

TobiHartmann (approved):
Looks reasonable.

vnkozlov (approved):
Good.

veresov:
Thanks for the reviews!
/integrate

PR #75, opened 2022-06-27
8289223: Canonicalize header ids in foreign API javadocs

JornVernee:
https://github.com/openjdk/jdk/pull/8817 added a button to copy a link to a section of javadoc to the clipboard. For the copy button to appear, the header needs to have an `id`.

This cleanup PR canonicalizes all header ids in the java.lang.foreign package to the preferred (non-legacy) style, and adds ids in places where they are missing as well.

In accordance with the Late-Enhancement Request Process [1], this is a `noreg-doc` documentation only change, which does not require an enhancement request.

[1]: https://openjdk.org/jeps/3#Late-Enhancement-Request-Process

---------
### Progress

### Issue
 * [JDK-8289223](https://bugs.openjdk.org/browse/JDK-8289223): Canonicalize header ids in foreign API javadocs

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/75/head:pull/75` \
`$ git checkout pull/75`

`$ git checkout pull/75` \
`$ git pull https://git.openjdk.org/jdk19 pull/75/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 75`

`$ git pr show -t 75`

</details>

<a href="https://git.openjdk.org/jdk19/pull/75.diff">https://git.openjdk.org/jdk19/pull/75.diff</a>

</details>

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/package-info.java
+++ b/src/java.base/share/classes/java/lang/foreign/package-info.java
@@ -27,7 +27,7 @@
 /**
  * <p>Provides low-level access to memory and functions outside the Java runtime.
  *
- * <h2>Foreign memory access</h2>
+ * <h2 id="fma">Foreign memory access</h2>
  *
  * <p>
  * The main abstraction introduced to support foreign memory access is {@link java.lang.foreign.MemorySegment}, which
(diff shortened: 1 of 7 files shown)


src/java.base/share/classes/java/lang/foreign/package-info.java
```diff
@@ -27,7 +27,7 @@
 /**
  * <p>Provides low-level access to memory and functions outside the Java runtime.
  *
- * <h2>Foreign memory access</h2>
+ * <h2 id="fma">Foreign memory access</h2>
```
mcimadamore, line 30: I would refrain from adding `id` if the section is never referred to.

JornVernee: The idea is that a user might want to link to this section of the javadoc (using the new copy button). For that to be possible, the section needs an `id`. That's why I've added one to all headers.

mcimadamore: I see what you mean - at the same time I don't see being done anywhere else in JDK, so perhaps can wait until we settle on a more uniform solution?

PR #81, opened 2022-06-27
8288854: getLocalGraphicsEnvironment() on for multi-screen setups throws exception NPE

mrserb:
This is the fix for a copy-paste error. The fix JDK-8076313 replaced the usage of the "screens" array from the parent class to the "devices" where the list of devices is now maintained. Since "screens" array is never used nor initialized its usage caused an NPE. That check was copied as-is, while it should use the actual number of screen devices requested early in that method.

The bug is rarely reproduced because in single screen configuration the main screen is usually 0, and in the multiscreen configuration Xinerama is usually active so the main screen is also 0 => the second part of the "if" statement is not executed.

I have validated the fix by the SwingSet, I also executed desktop tests in that config and found that even though this particular bug is fixed we still have many issues there, around ~100 tests failed.

---------
### Progress

### Issue
 * [JDK-8288854](https://bugs.openjdk.org/browse/JDK-8288854): getLocalGraphicsEnvironment() on for multi-screen setups throws exception NPE

### Reviewers
 * [Alexander Zvegintsev](https://openjdk.org/census#azvegint) (@azvegint - **Reviewer**)
 * [Alexey Ivanov](https://openjdk.org/census#aivanov) (@aivanov-jdk - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/81/head:pull/81` \
`$ git checkout pull/81`

`$ git checkout pull/81` \
`$ git pull https://git.openjdk.org/jdk19 pull/81/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 81`

`$ git pr show -t 81`

</details>

<a href="https://git.openjdk.org/jdk19/pull/81.diff">https://git.openjdk.org/jdk19/pull/81.diff</a>

</details>

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/X11GraphicsEnvironment.java
+++ b/src/java.desktop/unix/classes/sun/awt/X11GraphicsEnvironment.java
@@ -236,7 +236,7 @@ public final class X11GraphicsEnvironment extends SunGraphicsEnvironment {
             throw new AWTError("no screen devices");
         }
         int index = getDefaultScreenNum();
-        mainScreen = 0 < index && index < screens.length ? index : 0;
+        mainScreen = 0 < index && index < numScreens ? index : 0;
 
         for (int id = 0; id < numScreens; ++id) {
             devices.put(id, old.containsKey(id) ? old.remove(id) :


azvegint (approved):
Looks good.

mrserb:
> BTW why this PR is raised against JDK19 and not mainline?

Don't all fixes from JDK19 have merged to the mainline "automatically"?

mrserb:
> > Don't all fixes from JDK19 have merged to the mainline "automatically"?
> 
> I worried that [JDK19](https://openjdk.org/projects/jdk/19/) are in [RDP1](https://openjdk.org/jeps/3#rdp-1) phase right now.

This is the phase when all fixes are merged to the mainline.

From your links above:
" In RDP 1, if the bug is targeted and if time permits, develop and integrate a fix when ready." 

| | Fix | Drop |
-- | -- | -- |
| RDP1 | Current P1–P3 | All P4–P5 |
|           |  Targeted P1–P3 if time |  Targeted P1–P3|
|           |  P1–P5 doc/test bugs |  Targeted P1–P3|

Since the small fix for a recent regression is ready I do not see a reason to skip it?

aivanov-jdk (approved):

azvegint:
> Since the small fix for a recent regression is ready I do not see a reason to skip it?

OK, thanks for clarification.

PR #84, opened 2022-06-28
8289196: Pattern domination not working properly for record patterns

lahodaj:
When checking that a pattern in a switch case is not dominated by a preceding pattern, handling for record pattern is mostly missing. This patch is attempting to fix that, and as it requires a bit more code, it moves the check into `Check`.

---------
### Progress

### Issue
 * [JDK-8289196](https://bugs.openjdk.org/browse/JDK-8289196): Pattern domination not working properly for record patterns

### Reviewers
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - **Reviewer**) ⚠️ Review applies to [220fcc12](https://git.openjdk.org/jdk19/pull/84/files/220fcc12606d35abfa4467dd877bbfa5987b3872)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/84/head:pull/84` \
`$ git checkout pull/84`

`$ git checkout pull/84` \
`$ git pull https://git.openjdk.org/jdk19 pull/84/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 84`

`$ git pr show -t 84`

</details>

<a href="https://git.openjdk.org/jdk19/pull/84.diff">https://git.openjdk.org/jdk19/pull/84.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
@@ -4373,6 +4373,95 @@ public class Check {
         }
     }
 
+    void checkSwitchCaseLabelDominated(List<JCCase> cases) {
+        List<JCCaseLabel> caseLabels = List.nil();
+        for (List<JCCase> l = cases; l.nonEmpty(); l = l.tail) {
+            JCCase c = l.head;
+            for (JCCaseLabel label : c.labels) {
+                if (label.hasTag(DEFAULTCASELABEL) || TreeInfo.isNullCaseLabel(label)) {
+                    continue;
+                }
+                Type currentType = labelType(label);
+                for (JCCaseLabel testCaseLabel : caseLabels) {
+                    Type testType = labelType(testCaseLabel);
+                    if (types.isSubtype(currentType, testType) &&
+                        !currentType.hasTag(ERROR) && !testType.hasTag(ERROR)) {
+                        //the current label is potentially dominated by the existing (test) label, check:
+                        boolean dominated = false;
+                        if (label instanceof JCConstantCaseLabel) {
+                            dominated |= !(testCaseLabel instanceof JCConstantCaseLabel);
+                        } else if (label instanceof JCPatternCaseLabel patternCL &&
+                                   testCaseLabel instanceof JCPatternCaseLabel testPatternCaseLabel &&
+                                   TreeInfo.unguardedCaseLabel(testCaseLabel)) {
+                            dominated = patternDominated(testPatternCaseLabel.pat,
+                                                         patternCL.pat);
+                        }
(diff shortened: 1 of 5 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
```diff
@@ -4373,6 +4373,95 @@ void checkSwitchCaseStructure(List<JCCase> cases) {
...
+            });
+        }
+        private boolean patternDominated(JCPattern existingPattern, JCPattern currentPattern) {
+            Type existingPatternType = types.erasure(existingPattern.type);
+            Type currentPatternType = types.erasure(currentPattern.type);
+            if (existingPatternType.isPrimitive() ^ currentPatternType.isPrimitive()) {
+                return false;
+            }
+            if (existingPatternType.isPrimitive()) {
+                if (!types.isSameType(existingPatternType, currentPatternType)) {
```
vicente-romero-oracle, line 4423: note: shouldn't we return true if both are primitive and they are the same type? I mean this method seems to returns the right thing in that case but it is doing some extra analysis that doesn't seem to be necessary IMO

vicente-romero-oracle:
shouldn't this test case be rejected too:

```
class TestRecordPattern {
    int testRecordPatternsDominated2() {
        record R(int a) {}
        Object o = null;
        switch (o) {
            //case R(int a): return 1;
            case R(int a) when guard(): return -1;
            case R(int a): return 1;
            default: return 0;
        }
    }

    boolean guard() {
        return false;
    }
}
```

http://cr.openjdk.java.net/~gbierman/jep427+405/jep427+405-20220601/specs/patterns-switch-record-patterns-jls.html#jls-14.30.3 doesn't mention anything about guards if I understand correctly

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
lahodaj: We could, but I am not sure if that's not "too clever". Returning `true` when the type is same is based on the fact that only binding patterns (type test patterns) can be of primitive types, so the tests below will always return `true`. I can do that if strongly preferred, but I somewhat like the separation of concerns, where we verify the types here, and then the structure is checked below, regardless of the pattern's type.

lahodaj:
> nit: we could probably add this test case for completeness:
> 
> ```
> class Test {
>     int test() {
>         record R(int a) {}
>         Object o = null;
>         switch (o) {
>             case R(int a) when true: return -1;
>             case R(int a): return 1;
>         }
>     }
> }
> ```

Thanks, done.

vicente-romero-oracle (approved):
lgtm

PR #88, opened 2022-06-28
8287847: Fatal Error when suspending virtual thread after it has terminated

plummercj:
This fixes a bug in the debug agent when there is a request to suspend a virtual thread that has already terminated. The issue was that unless the debug agent was currently under a "suspend all", it would not properly put the virtual thread on the `otherThreads` list, and instead added it to `runningVThreads`. This meant at the end of `insertThread()` the following code tried to do a JVMTI `SetThreadLocalStorage`, which can't be done on a terminated thread.

```
        if (list != &otherThreads) {
            setThreadLocalStorage(node->thread, (void*)node);
        }
```

---------
### Progress

### Issue
 * [JDK-8287847](https://bugs.openjdk.org/browse/JDK-8287847): Fatal Error when suspending virtual thread after it has terminated

### Reviewers
 * [Alan Bateman](https://openjdk.org/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to [1ff59060](https://git.openjdk.org/jdk19/pull/88/files/1ff59060b9f0b50ceb83a6e36195c707a09f41c2)
 * [Serguei Spitsyn](https://openjdk.org/census#sspitsyn) (@sspitsyn - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/88/head:pull/88` \
`$ git checkout pull/88`

`$ git checkout pull/88` \
`$ git pull https://git.openjdk.org/jdk19 pull/88/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 88`

`$ git pr show -t 88`

</details>

<a href="https://git.openjdk.org/jdk19/pull/88.diff">https://git.openjdk.org/jdk19/pull/88.diff</a>

</details>

### DIFF
--- a/src/jdk.jdwp.agent/share/native/libjdwp/threadControl.c
+++ b/src/jdk.jdwp.agent/share/native/libjdwp/threadControl.c
@@ -413,10 +413,10 @@ insertThread(JNIEnv *env, ThreadList *list, jthread thread)
                 EXIT_ERROR(error, "getting vthread state");
             }
             if ((vthread_state & JVMTI_THREAD_STATE_ALIVE) == 0) {
-              // Thread not alive so put on otherThreads list instead of runningVThreads.
-              // It might not have started yet or might have terminated. Either way,
-              // otherThreads is the place for it.
-              list = &otherThreads;
+                // Thread not alive so put on otherThreads list instead of runningVThreads.
+                // It might not have started yet or might have terminated. Either way,
+                // otherThreads is the place for it.
+                list = &otherThreads;
             }
             if (suspendAllCount > 0) {
                 // Assume the suspendAllCount, just like the regular thread case above.
--- a/src/jdk.jdwp.agent/share/native/libjdwp/threadControl.c
+++ b/src/jdk.jdwp.agent/share/native/libjdwp/threadControl.c
@@ -412,6 +412,12 @@ insertThread(JNIEnv *env, ThreadList *list, jthread thread)
             if (error != JVMTI_ERROR_NONE) {
                 EXIT_ERROR(error, "getting vthread state");
             }
+            if ((vthread_state & JVMTI_THREAD_STATE_ALIVE) == 0) {
+              // Thread not alive so put on otherThreads list instead of runningVThreads.
+              // It might not have started yet or might have terminated. Either way,
+              // otherThreads is the place for it.
+              list = &otherThreads;
+            }
             if (suspendAllCount > 0) {
                 // Assume the suspendAllCount, just like the regular thread case above.
                 node->suspendCount = suspendAllCount;
@@ -419,7 +425,6 @@ insertThread(JNIEnv *env, ThreadList *list, jthread thread)
                     // If state == 0, then this is a new vthread that has not been started yet.
                     // Need to suspendOnStart in that case, just like the regular thread case above.
                     node->suspendOnStart = JNI_TRUE;
-                    list = &otherThreads; // Put on otherThreads list instead of runningVThreads
                 }
             }
             if (vthread_state != 0) {
--- a/test/jdk/com/sun/jdi/SuspendAfterDeath.java
+++ b/test/jdk/com/sun/jdi/SuspendAfterDeath.java
@@ -22,7 +22,7 @@
  */
 
 /**
- * @test
+ * @test id=platform_thread
  * @bug 8287847
  * @summary Test suspending a thread after it has terminated.
  * @enablePreview
@@ -31,6 +31,18 @@
  * @run compile SuspendAfterDeath.java
  * @run main/othervm SuspendAfterDeath
  */
+
+/**
+ * @test id=virtual_thread
+ * @bug 8287847
+ * @summary Test suspending a thread after it has terminated.
+ * @enablePreview
+ * @requires vm.continuations
+ * @run build TestScaffold VMConnection TargetListener TargetAdapter
+ * @run compile SuspendAfterDeath.java
+ * @run main/othervm SuspendAfterDeath Virtual
+ */
+
 import com.sun.jdi.*;
 import com.sun.jdi.event.*;
 import com.sun.jdi.request.*;
@@ -60,13 +72,23 @@ class SuspendAfterDeathTarg {
 public class SuspendAfterDeath extends TestScaffold {
     private volatile ThreadReference thread;
     private volatile boolean breakpointReached;
+    private static boolean useVirtualThread = false;
 
-    SuspendAfterDeath(String args[]) {
-        super(args);
+    SuspendAfterDeath() {
+        super(new String[0]); // no args to pass along to debuggee
     }
 
     public static void main(String[] args) throws Exception {
-        new SuspendAfterDeath(args).startTests();
+        if (args.length == 1) {
+            if ("Virtual".equals(args[0])) {
+                useVirtualThread = true; // see connect() below for how this is handled
+            } else {
+                throw new RuntimeException("SuspendAfterDeath: invalid argument: " + args[0]);
+            }
+        } else if (args.length != 0) {
+            throw new RuntimeException("SuspendAfterDeath: incorrect number of arguments: " + args.length);
+        }
+        new SuspendAfterDeath().startTests();
     }
 
     @Override
@@ -92,8 +114,8 @@ public class SuspendAfterDeath extends TestScaffold {
 
     @Override
     public void connect(String args[]) {
-        String mainWrapper = System.getProperty("main.wrapper");
-        if ("Virtual".equals(mainWrapper)) {
+        if (useVirtualThread) {
+            /* Append the "Virtual" argument to the arguments used for the debuggee. */
             List<String> argList = new ArrayList(Arrays.asList(args));
             argList.add("Virtual");
             args = argList.toArray(args);
--- a/test/jdk/com/sun/jdi/SuspendAfterDeath.java
+++ b/test/jdk/com/sun/jdi/SuspendAfterDeath.java
@@ -59,7 +59,7 @@ class SuspendAfterDeathTarg {
 
 public class SuspendAfterDeath extends TestScaffold {
     private volatile ThreadReference thread;
-    private volatile boolean breakpointReached = false;
+    private volatile boolean breakpointReached;
 
     SuspendAfterDeath(String args[]) {
         super(args);
--- a/test/jdk/com/sun/jdi/SuspendAfterDeath.java
+++ b/test/jdk/com/sun/jdi/SuspendAfterDeath.java
@@ -94,9 +94,9 @@ public class SuspendAfterDeath extends TestScaffold {
     public void connect(String args[]) {
         String mainWrapper = System.getProperty("main.wrapper");
         if ("Virtual".equals(mainWrapper)) {
-            List argList = new ArrayList(Arrays.asList(args));
+            List<String> argList = new ArrayList(Arrays.asList(args));
             argList.add("Virtual");
-            args = (String[]) argList.toArray(args);
+            args = argList.toArray(args);
         }
         super.connect(args);
     }
--- a/test/jdk/com/sun/jdi/SuspendAfterDeath.java
+++ b/test/jdk/com/sun/jdi/SuspendAfterDeath.java
@@ -36,7 +36,6 @@ import com.sun.jdi.event.*;
 import com.sun.jdi.request.*;
 import java.util.*;
 
-
 class SuspendAfterDeathTarg {
     static final String THREAD_NAME = "duke";
 
@@ -61,7 +60,7 @@ class SuspendAfterDeathTarg {
 public class SuspendAfterDeath extends TestScaffold {
     private volatile ThreadReference thread;
     private volatile boolean breakpointReached = false;
-    
+
     SuspendAfterDeath(String args[]) {
         super(args);
     }
@@ -85,7 +84,7 @@ public class SuspendAfterDeath extends TestScaffold {
         System.out.println("Breakpoint, thread=" + eventThread);
         if (thread == null) {
             failure("FAILED: got Breakpoint event before ThreadDeath event.");
-        }            
+        }
         breakpointReached = true;
         /* Suspend the thread. This is being done after the thread has exited. */
         thread.suspend();
--- a/test/jdk/com/sun/jdi/SuspendAfterDeath.java
+++ b/test/jdk/com/sun/jdi/SuspendAfterDeath.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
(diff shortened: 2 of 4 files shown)


test/jdk/com/sun/jdi/SuspendAfterDeath.java
```diff
@@ -0,0 +1,135 @@
...
+            thread = Thread.ofPlatform().name(THREAD_NAME).start(() -> { });
+        }
+        thread.join();
+        done();
+    }
+}
+
+public class SuspendAfterDeath extends TestScaffold {
+    private volatile ThreadReference thread;
+    private volatile boolean breakpointReached = false;
```
AlanBateman, line 62: The volatile-write to initialise this to false is not needed here.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
```diff
@@ -0,0 +1,135 @@
...
+        thread.suspend();
+    }
+
+    @Override
+    public void connect(String args[]) {
+        String mainWrapper = System.getProperty("main.wrapper");
+        if ("Virtual".equals(mainWrapper)) {
+            List argList = new ArrayList(Arrays.asList(args));
+            argList.add("Virtual");
+            args = (String[]) argList.toArray(args);
```
AlanBateman, line 99: The raw type and casting caught my addition here. Here's something more succulent if you'd like:
```
args = Stream.concat(Stream.of(args), Stream.of("Virtual"))
                .toArray(String[]::new);
```

test/jdk/com/sun/jdi/SuspendAfterDeath.java
```diff
@@ -0,0 +1,135 @@
...
+            failure("FAILED: never got Breakpoint event for target thread.");
+        }
+
+        if (!testFailed) {
+            println("SuspendAfterDeath: passed");
+        } else {
+            throw new Exception("SuspendAfterDeath: failed");
+        }
+    }
+}
```
AlanBateman, line 135: I wonder if we need to add `@run main/othervm -Dmain.wrapper=Virtual SuspendAfterDeath` to the test description so that running the jdk_jdi test group will exercise the issue. As it stands, I think it would require. test run with -Dmain.wrapper=Virtual to exercise this code.

src/jdk.jdwp.agent/share/native/libjdwp/threadControl.c
```diff
@@ -412,14 +412,19 @@ insertThread(JNIEnv *env, ThreadList *list, jthread thread)
             if (error != JVMTI_ERROR_NONE) {
                 EXIT_ERROR(error, "getting vthread state");
             }
+            if ((vthread_state & JVMTI_THREAD_STATE_ALIVE) == 0) {
+              // Thread not alive so put on otherThreads list instead of runningVThreads.
+              // It might not have started yet or might have terminated. Either way,
+              // otherThreads is the place for it.
+              list = &otherThreads;
```
AlanBateman, line 419: Minor nit but the native code here uses 4-space indentation.

src/jdk.jdwp.agent/share/native/libjdwp/threadControl.c
```diff
@@ -412,14 +412,19 @@ insertThread(JNIEnv *env, ThreadList *list, jthread thread)
             if (error != JVMTI_ERROR_NONE) {
                 EXIT_ERROR(error, "getting vthread state");
             }
+            if ((vthread_state & JVMTI_THREAD_STATE_ALIVE) == 0) {
+              // Thread not alive so put on otherThreads list instead of runningVThreads.
+              // It might not have started yet or might have terminated. Either way,
+              // otherThreads is the place for it.
```
AlanBateman, line 418: If a terminated thread is added to otherThreads then will be it eventually removed?

plummercj: Yes, I had the same question. otherThreads is swept every time any thread is resumed. This code really isn't much different than what is done for platform threads, except that the detection that it should go on otherThreads is done earlier on commonSuspend(). The code differences arise from the differences in how platform threads and virtual threads are tracked. I hope to unify them someday, but for now don't want virtual thread support to risk destabilizing platform thread support.

src/jdk.jdwp.agent/share/native/libjdwp/threadControl.c
plummercj: fixed

test/jdk/com/sun/jdi/SuspendAfterDeath.java
plummercj: Actually I was a bit confused as to why thread was declared volatile, so I just followed the pattern. Maybe you can explain since you wrote it.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
plummercj: That came from TestScaffold.startUp():
```
    protected void startUp(String targetName) {
        List argList = new ArrayList(Arrays.asList(args));
        argList.add(targetName);
        println("run args: " + argList);
        connect((String[]) argList.toArray(args));
        waitForVMStart();
    }
```
TBH I simply find myself staring at Streams code like that for way too long before I figure out what it is trying to do. If you use/read code like that a lot, it's probably not an issue, but I don't, and prefer something more straight forward, even if it is not nearly as elegant. Someone should add `String[] Arrays.appendToArray(String[] arr, String s)`, or even `String[] Arrays.concat(String[] s1, String[] s2)` would do.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
plummercj: Yes, it requires `-Dmain.wrapper=Virtual`, and that was intentional. My thinking was that we don't have any com/sun/jdi tests that do any virtual thread testing unless run with `-Dmain.wrapper=Virtual`, so why should this test be any different? If I made it do virtual thread testing without `-Dmain.wrapper=Virtual`, then the `-Dmain.wrapper=Virtual` testing becomes redundant.

AlanBateman: The test uses a platform thread by default so it's not testing the issue in JDK-8287847. The suggestion is to have it re-run with the system property set, like this:
```
@run main/othervm SuspendAfterDeath 
@run main/othervm -Dmain.wrapper=Virtual SuspendAfterDeath 
```
I accept this may be different to the existing com/sun/jdi tests but they pre-date virtual threads.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
AlanBateman: II'm just pointing out that initialising breakpointReached to false is unnecessary as it's the default value.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
AlanBateman: Whatever you are comfortable with but I think we should add the raw type and avoid the cast.

plummercj: fixed

test/jdk/com/sun/jdi/SuspendAfterDeath.java
plummercj: but do either of these fields need to be volatile?

test/jdk/com/sun/jdi/SuspendAfterDeath.java
plummercj: The @run suggestion doesn't work. There are two issues. First, it does not trigger jtreg to use TestScaffold.main(), which is what normally happens when you use TEST_VM_OPTS="-Dmain.wrapper=Virtual". So the end result is that the test is run with a platform thread.

The other issue is that if the test run is done with TEST_VM_OPTS="-Dmain.wrapper=Virtual", then TestScaffold.main() cannot handle any arguments added to the @run other than the main class name. You end up with CNFE for the first argument that appears after "SuspendAfterDeath".

The above issues aside, I still don't understand why we would want to do any virtual thread testing with com/sun/jdi tests when not using TEST_VM_OPTS="-Dmain.wrapper=Virtual". It's not how we've tested JDI support for virtual thread in the past. We've always relied on explicit virtual thread test runs using TEST_VM_OPTS="-Dmain.wrapper=Virtual". Same is true of the nsk jdi, jdb, and jdwp tests. If you really want this test handled differently because it was written to test a specific bug, then it probably does not belong in com/sun/jdi, which means it can't rely on TestScaffold.

One other thing is that you don't really need -Dmain.wrapper=Virtual to make this test do the testing on a virtual thread. It can just be the default mode like you used to have when you first wrote it. However, I ran into problems with that when also using  TEST_VM_OPTS="-Dmain.wrapper=Virtual" due to conflicts in argument setup between the test and TestScaffold. However, with the changes now in parseArgs() to check test.enable.preview, it might not be necessary for the test itself to add --enable-preview, so maybe that conflict will go away. But I still don't think this is the approach we should take (see my previous paragraph).

AlanBateman: The testing in OpenJDK does not run these tests with -Dmain.wrapper=Virtual so this is why this test is not testing the scenario that this bug is about.  We can make it work and have it run twice, or we can rename it to something like "SuspendVirtualThreadAfterDeath" and drop the testing with a platform thread. In time I expect there will be more tests in the jdk_jdi test group that exercise virtual threads.




It can be made to work but I would be okay with dropping the testing of platform thread from this test. 


The reason for asking for changes here is the test, as presented, doesn't exercise virtual threads with any of the tiers defined

plummercj: I've updated the test to always run in both virtual thread and platform thread mode. The test itself does not care about the `-Dmain.wrapper=Virtual` setting and ignores it. TestScaffold does care and will run the main test thread as a virtual thread if `-Dmain.wrapper=Virtual` is used, but the test itself will spawn a virtual thread or platform thread independent of the wrapper setting.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test id=platform_thread
+ * @bug 8287847
+ * @summary Test suspending a thread after it has terminated.
```
AlanBateman, line 27: Minor comment but this could say "suspend a platform thread" (and "virtual thread" for the other summary).

AlanBateman (approved):

AlanBateman (approved):

plummercj:
I decided to check the status of the target thread when the breakpoint was reached, and as expected it was incorrect in the virtual thread case due to [JDK-8288703](https://bugs.openjdk.org/browse/JDK-8288703). The status was THREAD_STATUS_UNKNOWN instead of THREAD_STATUS_ZOMBIE. I applied Serguei's JVMTI fix and that resolved the problem. Once Serguei pushes #102 I'll merge, verify my test changes, and then commit them for further review.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
```diff
@@ -0,0 +1,157 @@
...
+        } else {
+            thread = Thread.ofPlatform().name(THREAD_NAME).start(() -> { });
+        }
+        thread.join();
+        done();
+    }
+}
+
+public class SuspendAfterDeath extends TestScaffold {
+    private volatile ThreadReference thread;
```
sspitsyn, line 73: One minor naming suggestion:
The name `thread` is too generic. What about something like `targetThread` or `testedThread`?

test/jdk/com/sun/jdi/TestScaffold.java
```diff
@@ -969,6 +973,8 @@ public static void main(String[] args) throws Throwable {
                     tg.uncaughtThrowable = error;
                 }
             });
+            Thread.currentThread().setName(OLD_MAIN_THREAD_NAME);
```
sspitsyn, line 976: Just want to understand what is this thread naming is about.
Is it just to properly name the threads, or there is some other reason for this?
Also, why OLD_MAIN_THREAD_NAME is a little bit strange and why old?

sspitsyn (approved):

test/jdk/com/sun/jdi/SuspendAfterDeath.java
plummercj: Ok. I'll rename to targetThread.

test/jdk/com/sun/jdi/TestScaffold.java
plummercj: The "main" debuggee thread is always launched as a platform thread. There's no choice in that since it is simply the main application thread. When using the virtual thread wrapper, the "main" thread will spawn a virtual thread to run the debuggee on. By default virtual threads have no name. For the nsk debugger tests, "main" was renamed to "old-m-a-i-n", and the new debuggee virtual thread named "main". For some reason that was never done for the com/sun/jdi tests. I only made this change because in `breakpointReached()` of the test it prints out the thread, and I noticed the name was empty.  I could leave this change out if you want and file a separate CR for it. The test should still run ok without it.

sspitsyn: Thank you for explaining. Just wanted to understand the motivation.

test/jdk/com/sun/jdi/SuspendAfterDeath.java
plummercj: Done.

plummercj:
Serguei has pushed his JVMTI fix for the virtual thread state and I have pushed the test changes to verify the target thread's state. Please review. Thanks

sspitsyn (approved):

AlanBateman (approved):

sspitsyn:
Chris,
I see the test SelfSuspendDisablerTest with my update failed in your sanity checks.

The problem is:
 ```
71     public static void main(String argv[]) throws Exception {
 72         Thread t1 = Thread.ofPlatform().factory().newThread(() -> {
 73             testJvmtiThreadState(Thread.currentThread(), RUNNABLE);
 74             selfSuspend();                       <== It does a self suspend here!!
 75         });
 76         Thread t2 = Thread.ofVirtual().factory().newThread(() -> {
 77             testJvmtiThreadState(Thread.currentThread(), RUNNABLE);
 78             while(!isSuspended(t1)) {
 79               Thread.yield();
 80             }
 81             Thread.yield(); // provoke unmount
 82 
 83             testJvmtiThreadState(t1, SUSPENDED);
 84 
 85             resume(t1);
 86 
 87             testJvmtiThreadState(t1, RUNNABLE);  <== This is wrong because of line 74!!
 88 
 89             suspendAllVirtualThreads();
 90         });
```
The line 87 has to be removed.
Could, you remove the lines 87-88 as part of your PR?
It’d be the simplest way to handle it.

sspitsyn (approved):
In fact, webrev v9 does not have any changes.
Is it a merge with JDK 19?

plummercj:
> In fact, webrev v9 does not have any changes.
> Is it a merge with JDK 19?

Yes, just a merge with the JDK 19 master branch.

PR #90, opened 2022-06-29
8289302: Restore CriticalJNINatives

coleenp:
This change restores the code in the JVM that implemented the internal JavaCritical or CriticalJNINatives functionality.  Customers are using this feature and there's no replacement yet for it.
This change is a backout and merge of jDK-8258192: Obsolete the CriticalJNINatives flag.
This change also adds the deprecation message that was missed, which I checked is okay with the customer.

Tested with tiers1-4.

---------
### Progress

### Issue
 * [JDK-8289302](https://bugs.openjdk.org/browse/JDK-8289302): Restore CriticalJNINatives

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/90/head:pull/90` \
`$ git checkout pull/90`

`$ git checkout pull/90` \
`$ git pull https://git.openjdk.org/jdk19 pull/90/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 90`

`$ git pr show -t 90`

</details>

<a href="https://git.openjdk.org/jdk19/pull/90.diff">https://git.openjdk.org/jdk19/pull/90.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
+++ b/src/hotspot/cpu/aarch64/sharedRuntime_aarch64.cpp
@@ -977,6 +977,79 @@ static void restore_args(MacroAssembler *masm, int arg_count, int first_arg, VMR
   }
 }
 
+// Unpack an array argument into a pointer to the body and the length
+// if the array is non-null, otherwise pass 0 for both.
+static void unpack_array_argument(MacroAssembler* masm, VMRegPair reg, BasicType in_elem_type, VMRegPair body_arg, VMRegPair length_arg) { Unimplemented(); }
+
+
+class ComputeMoveOrder: public StackObj {
+  class MoveOperation: public ResourceObj {
+    friend class ComputeMoveOrder;
+   private:
+    VMRegPair        _src;
+    VMRegPair        _dst;
+    int              _src_index;
+    int              _dst_index;
+    bool             _processed;
+    MoveOperation*  _next;
+    MoveOperation*  _prev;
+
+    static int get_id(VMRegPair r) { Unimplemented(); return 0; }
+
+   public:
+    MoveOperation(int src_index, VMRegPair src, int dst_index, VMRegPair dst):
+      _src(src)
+    , _dst(dst)
+    , _src_index(src_index)
+    , _dst_index(dst_index)
+    , _processed(false)
+    , _next(NULL)
+    , _prev(NULL) { Unimplemented(); }
+
+    VMRegPair src() const              { Unimplemented(); return _src; }
+    int src_id() const                 { Unimplemented(); return 0; }
+    int src_index() const              { Unimplemented(); return 0; }
+    VMRegPair dst() const              { Unimplemented(); return _src; }
+    void set_dst(int i, VMRegPair dst) { Unimplemented(); }
+    int dst_index() const              { Unimplemented(); return 0; }
+    int dst_id() const                 { Unimplemented(); return 0; }
+    MoveOperation* next() const        { Unimplemented(); return 0; }
+    MoveOperation* prev() const        { Unimplemented(); return 0; }
+    void set_processed()               { Unimplemented(); }
+    bool is_processed() const          { Unimplemented(); return 0; }
+
+    // insert
+    void break_cycle(VMRegPair temp_register) { Unimplemented(); }
+
+    void link(GrowableArray<MoveOperation*>& killer) { Unimplemented(); }
+  };
+
+ private:
+  GrowableArray<MoveOperation*> edges;
+
+ public:
+  ComputeMoveOrder(int total_in_args, VMRegPair* in_regs, int total_c_args, VMRegPair* out_regs,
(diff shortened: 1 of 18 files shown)


robehn:
I'm not convinced.

The example provided for a short native call that is a lot slower is:
clock_gettime(CLOCK_REALTIME)

This is vDSO syscall, meaning it is just reading mapped memory.
If performance was essential why is the user not mapping the vDSO memory directly into Java, thus plainly reading without even needing to do a native method call?

coleenp:
I assume it's because the user just wants to call clock_gettime(CLOCK_REALTIME) using the os API provided.

coleenp:
Panama had this sort of functionality and didn't want to keep it. Hopefully they can answer.  They're working on satisfying the needs of this sort of customers, but until then, this is something that will prevent them from moving forward.

AlanBateman:
I would have liked to have seen more from the person that came to jdk-dev and panama-dev to complain about this. In particular, the mails mention replacing parts of NIO which suggests to me that this might be a JDK fork, or maybe the undocumented CriticalJNINative in conjunction with other dependences on JDK internals? It came across as completely unsupportable and I think engagement on panama-dev would be better before making any decision to bring this back.

mcimadamore:
> What is the panama take on this?

At some point Panama provided support for "trivial" calls, e.g. native calls whose Java -> native transition was elided. In more recent iterations of the API we ended up leaving that "feature" on the side: it only provided very limited boost (e.g. ~4-5 ns) and at great cost in terms of risks (e.g. if the native call upcalls back to Java a JVM crash is unavoidable).

As pointed out elsewhere [1, 2], a much better way to support this kind of use cases is through new intrinsics, as JVM intrinsics would avoid the typical costs associated with a native call (critical or not).

[1] - https://mail.openjdk.org/pipermail/panama-dev/2022-June/017048.html
[2] - https://mail.openjdk.org/pipermail/panama-dev/2022-June/017050.html

kirillp:
I can confirm very significant performance drop. We have been using JavaCritical for OpenGL calls, for FreeType & HarfBuzz calls,
for libPNG interop, for TurboJPEG interop. Basically for everything and it makes a very significant performance gain 

Performance numbers were investigated during GraalVM evaluations here: https://github.com/oracle/graal/issues/2873
You can find all source code there
And the result is  JavaCritical is about 6x-7x faster:
```
  This is my results, numbers is time spent, lower better:
Jvm
  min critical: 84200
  min non critical: 558200

```

coleenp:
@kirillp Can you try your measurements with the current JDK 20 build with and without the option:
-XX:+UseSystemMemoryBarrier

robehn:
Note that if there is a safepoint or handshake without JavaCritical the JavaThread may be stopped for such on the back edge of the call. Which means if you measure the time from Java you also including time spend in safepoint/handshake in such case.

We measure using https://github.com/openjdk/jmh, which can control for such events.
Here are some example of our numbers:

########### BASELINE JDK17 ##########
Benchmark - Iterations: 100000000 total time:3446158970 ns ; Per op:34.4615897
Benchmark - Iterations: 100000000 total time:3481146574 ns ; Per op:34.81146574
Benchmark - Iterations: 100000000 total time:3432662217 ns ; Per op:34.32662217
########### CRITICAL JDK17 ##########
Benchmark - Iterations: 100000000 total time:2695102198 ns ; Per op:26.95102198
Benchmark - Iterations: 100000000 total time:2707413238 ns ; Per op:27.07413238
Benchmark - Iterations: 100000000 total time:2604335910 ns ; Per op:26.0433591
###########  UseSystemMemoryBarrier JDK17 ##########
Benchmark - Iterations: 100000000 total time:2930762287 ns ; Per op:29.30762287
Benchmark - Iterations: 100000000 total time:2874010646 ns ; Per op:28.74010646
Benchmark - Iterations: 100000000 total time:2817695430 ns ; Per op:28.1769543

Note that we accidentally optimized JavaCritical in JDK 17, so UseSystemMemoryBarrier should be on par with pre-JDK-17.

TheRealMDoerr:
`UseSystemMemoryBarrier` reduces the overhead of Java thread state transitions significantly (when returning from native and when calling JNI functions).
But there's one additional thing which should be checked regarding handling of primitive arrays. Critical JNI functions provide direct access to the arrays in the Java heap while the regular JNI functions need to use Get/Set<PrimitiveType>ArrayRegion (which is fast for small arrays or when copying is required anyway) or Get<PrimitiveType>ArrayElements&Release<PrimitiveType>ArrayElements (slower, but copies forth and back) or GetPrimitiveArrayCritical&ReleasePrimitiveArrayCritical (which can avoid copying costs, but require expensive interaction with GC). It's probably worth checking the JNI code for good selection of array access functions.

TheRealMDoerr:
Btw. performance of GetPrimitiveArrayCritical&ReleasePrimitiveArrayCritical could get improved for weak memory model platforms by relaxing memory ordering semantics of `Atomic::inc/dec` in object pinning (Shenandoah) and `GCLocker` handling (debug build, not sure if there's more to optimize) code.

PR #91, opened 2022-06-29
8289148: j.l.foreign.VaList::nextVarg call could throw IndexOutOfBoundsException or even crash the VM

JornVernee:
This patch changes all VaList implementations to throw `NoSuchElementException` when out of bounds reads occur on a VaList that is created using the Java builder API. The docs are updated accordingly.

For VaLists that are created from native addresses, we don't know their bounds, so we can not throw exceptions in that case.

Testing: `jdk_foreign` test suite on all platforms that implement VaList.

---------
### Progress
- [x] Change requires a CSR request to be approved

### Issues
 * [JDK-8289148](https://bugs.openjdk.org/browse/JDK-8289148): j.l.foreign.VaList::nextVarg call could throw IndexOutOfBoundsException or even crash the VM
 * [JDK-8289333](https://bugs.openjdk.org/browse/JDK-8289333): Specification of method j.l.foreign.VaList::skip deserves clarification
 * [JDK-8289156](https://bugs.openjdk.org/browse/JDK-8289156): j.l.foreign.VaList::skip call could throw java.lang.IndexOutOfBoundsException: Out of bound access on segment
 * [JDK-8289540](https://bugs.openjdk.org/browse/JDK-8289540): j.l.foreign.VaList::nextVarg call could throw IndexOutOfBoundsException or even crash the VM (**CSR**)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - **Reviewer**) ⚠️ Review applies to [ba132af5](https://git.openjdk.org/jdk19/pull/91/files/ba132af59831eb56671edf24d6233033a1046435)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/91/head:pull/91` \
`$ git checkout pull/91`

`$ git checkout pull/91` \
`$ git pull https://git.openjdk.org/jdk19 pull/91/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 91`

`$ git pr show -t 91`

</details>

<a href="https://git.openjdk.org/jdk19/pull/91.diff">https://git.openjdk.org/jdk19/pull/91.diff</a>

</details>


src/java.base/share/classes/java/lang/foreign/VaList.java
```diff
@@ -50,7 +51,14 @@
...
  * and any other type that fits into a {@code long}.
- *
+ * <h2 id="safety">Safety</h2>
+ * The {@code nextVarg} methods used to retrieve elements from a variable argument list
+ * will try to detect out-of-bounds reads on a best-effort basis.
+ * <p>
+ * In practice, whether this detection succeeds depends on the method used to create the variable argument list.
+ * Particularly, variable argument lists created using {@link #make(Consumer, MemorySession)} are able to detect out-of-bounds reads,
+ * while variable argument lists crearted using {@link #ofAddress(MemoryAddress, MemorySession)} are not.
+ * <p>
```
mcimadamore, line 61:
```suggestion
 * <h2>Safety considerations</h2>
 * A variable argument list can be thought of as a cursor into a memory segment which stores one or
 * more elements, with given layouts. The {@linkplain #nextVarg(ValueLayout.OfInt) methods} used to retrieve elements from
 * a variable argument list update the cursor position; as such it is possible for clients to access elements
 * outside the spatial bounds of the variable argument list.
 * <p>
 * A variable argument list attempts to detect out-of-bounds access on a best-effort basis.
 * Whether this detection succeeds depends on the factory method used to create the variable argument list:
 * <ul>
 *     <li>Variable argument lists created <em>safely</em>, using {@link #make(Consumer, MemorySession)} are built on top of
 *     native memory segments that are sized correctly. As such, they are capable of detecting out-of-bounds reads;</li>
 *     <li>Variable argument lists created <em>unsafely</em>, using {@link #ofAddress(MemoryAddress, MemorySession)} are built on top
 *     of segments whose size is unknown. As such they cannot detect out-of-bounds reads</li>
 * <p>```

File: src/java.base/share/classes/java/lang/foreign/VaList.java [hunk 2]
```diff
@@ -87,6 +96,7 @@ sealed public interface VaList extends Addressable permits WinVaList, SysVVaList
      * {@linkplain MemorySession#isAlive() alive}.
      * @throws WrongThreadException if this method is called from a thread other than the thread owning
      * the {@linkplain #session() session} associated with this variable argument list.
+     * @throws NoSuchElementException if an out-of-bounds read is detected.
```
mcimadamore, line 99: I'd insert a link to the safety section, where the link text is "out-of-bounds".

src/java.base/share/classes/java/lang/foreign/VaList.java
```diff
@@ -185,6 +199,8 @@ sealed public interface VaList extends Addressable permits WinVaList, SysVVaList
      * the JVM or, worse, silently result in memory corruption. Thus, clients should refrain from depending on
      * restricted methods, and use safe and supported functionalities, where possible.
      *
+     * @implNote variable argument lists created using this method can not detect out-of-bound reads.
```
mcimadamore, line 202: Links here too

src/java.base/share/classes/jdk/internal/foreign/abi/x64/sysv/SysVVaList.java
```diff
@@ -274,6 +288,22 @@ private Object read(Class<?> carrier, MemoryLayout layout, SegmentAllocator allo
         }
     }
 
+    private void checkRegAreaElement(MemoryLayout layout, TypeClass typeClass) {
```
mcimadamore, line 291: should this be `checkRegSaveAreaElement` ?

test/jdk/java/foreign/valist/VaListTest.java
```diff
@@ -815,4 +822,87 @@ static MethodHandle mh(VaListConsumer instance) {
             }
         }
     }
+
+    @DataProvider
+    public static Object[][] overflow() {
```
mcimadamore, line 827: nice test!

src/java.base/share/classes/java/lang/foreign/VaList.java
JornVernee: This seems a bit too much.

The class javadoc further up already describes a va list as "a stateful cursor used to iterate over a set of arguments". If that description is insufficient, I think it should be amended at that point.

Also, I don't think we have to go into describing how va lists returned by different factories are implemented (in fact, I think we should avoid that in order not to make accidental false promises when things change). It seems like noise next to the safety concerns (if someone really wants to know how the specified semantics are implemented, they can look at the code, or ask on the mailing list).

I'd suggest something simpler like this:

```suggestion
 * It is possible for clients to access elements outside the spatial bounds of a variable argument list. Variable argument list
 * implementations will try to detect out-of-bounds reads on a best-effort basis.
 * <p>
 * Whether this detection succeeds depends on the factory method used to create the variable argument list:
 * <ul>
 *     <li>Variable argument lists created <em>safely</em>, using {@link #make(Consumer, MemorySession)} are capable of detecting out-of-bounds reads;</li>
 *     <li>Variable argument lists created <em>unsafely</em>, using {@link #ofAddress(MemoryAddress, MemorySession)} are not capable of detecting out-of-bounds reads</li>
 * </ul>
 * <p>
```

I'm also fine with changing the section title to "Safety considerations".

JornVernee: i.e. I'd like to just specify the behavior, and avoid explaining why we have that behavior.

src/java.base/share/classes/java/lang/foreign/VaList.java
JornVernee: Good idea, I'll add links

src/java.base/share/classes/jdk/internal/foreign/abi/x64/sysv/SysVVaList.java
JornVernee: Okay, I can rename it.

src/java.base/share/classes/java/lang/foreign/VaList.java
JornVernee: I've updated the PR with this suggestion for now. Please let me know if it looks OK to you as well.

mcimadamore (approved):
Looks good!

PR #101, opened 2022-07-01
8289486: Improve XSLT XPath operators count efficiency

JoeWang-Java:
To improve efficiency, this patch moves the limit check to within the Lexer and reports any overlimit situation as soon as it happens.

Note the change in XPathParser: diff (and also webrevs) showed the whole error-report block was changed, the actual change was only placing the block to within the isOverLimit statement.

Test: java.xml tests passed.

---------
### Progress

### Issue
 * [JDK-8289486](https://bugs.openjdk.org/browse/JDK-8289486): Improve XSLT XPath operators count efficiency

### Reviewers
 * [Naoto Sato](https://openjdk.org/census#naoto) (@naotoj - **Reviewer**) ⚠️ Review applies to [330f86ce](https://git.openjdk.org/jdk19/pull/101/files/330f86ceeea8b5c60b9c05f402f1405debe0247d)
 * [Lance Andersen](https://openjdk.org/census#lancea) (@LanceAndersen - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/101/head:pull/101` \
`$ git checkout pull/101`

`$ git checkout pull/101` \
`$ git pull https://git.openjdk.org/jdk19 pull/101/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 101`

`$ git pr show -t 101`

</details>

<a href="https://git.openjdk.org/jdk19/pull/101.diff">https://git.openjdk.org/jdk19/pull/101.diff</a>

</details>


naotoj:
Hi Joe,
In the bug report, it reads:
```
It would be more efficient for the XSLT XPath impl to do the same as the XPath impl does and report any error earlier
```
So now the code works as once the first error was found, it immediately reports it. Would that result in a behavioral change, if there are multiple overflows co-exist? Would there be a possibility where a different error be thrown compared to the prior implementation?

JoeWang-Java:
Hi Naoto, 

Thanks for reviewing.

For counting, it's mainly the way the totalOpCount was accumulated (e.g. previously it was totalOpCount += opCount and now totalOpCount++). But since the limit on total ops is usually significantly larger than that of the count for one expression, this change won't have effect on how it was reported, that is, it won't be report before individual limit is reached. In that regard, the reporting sequence hasn't changed (that is, group -> individual count -> total). 

The other factor is that this feature is fairly new, the change therefore will unlikely affect any real-world applications.

The test cases, reporting group/individual/total errors in the order, continued passing, that's a good indicator as well.

naotoj (approved):
Thanks for the explanation. Looks good to me.

PR #107, opened 2022-07-04
8289601: MemorySegment.allocateUtf8String(String str) should be clarified for strings containing \0

JornVernee:
This PR updates the spec and implementation to throw an `IllegalArgumentException` when an attempt is made to convert a Java string containing null characters to a C string.

### DIFF
--- a/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
+++ b/src/java.base/share/classes/java/lang/foreign/SegmentAllocator.java
@@ -76,6 +76,7 @@ public interface SegmentAllocator {
      * into a new memory segment obtained by calling {@code this.allocate(str.length() + 1)}.
      * @param str the Java string to be converted into a C string.
      * @return a new native memory segment containing the converted C string.
+     * @throws IllegalArgumentException if the UTF-8 encoded bytes of the given string contain a {@code 0} byte.
      */
     default MemorySegment allocateUtf8String(String str) {
         Objects.requireNonNull(str);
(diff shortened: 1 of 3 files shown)


PaulSandoz (commented):
It's tempting to allow null/0 characters to pass through, truncating the string, given the rarity of such characters, but it could hide a nasty bug.

If performance becomes a really important issue we can make intrinsic and vectorize.

I think it is better to refer to the characters in the String, rather than refer to bytes of the UTF-8 encoding of that String e.g. throws if the string contains a null character (same for the exception message). We can explain in an API note why.

JimLaskey:
Is UTF-8 E0 C0 reasonable in this case? Or at least an option?

mcimadamore:
After some offline discussions Jorn and I have decided to leave things as they are and document the behavior. After all, there's no unsafety - and in some cases (e.g. concatenating multiple strings into a "flat" array whose elements are separated by terminators) it might even be desirable. The performance impact of checking all bytes and throwing is just too much: from 1.5 slower (with 100 chars) when using a standard allocator, up to 2.5x slower when using an optimized allocator. As a fallback, users could always resort to manual bulk copy, but the trade off would not be super obvious.

JornVernee:
I've tweaked the Javadoc. I realized it should also be tweaked for `MemorySegment::setUtf8String`.

PaulSandoz (approved):
That is indeed the simplest option, and the more preferable one given the performance cost of checking.

PR #95, opened 2022-06-30
8067757: Incorrect HTML generation for copied javadoc with multiple @throws tags

pavelrappo:
Please review this PR for JDK 19.

This PR fixes JDK-8067757. To understand what JDK-8067757 is about, you first need to understand how javadoc documents the fact that such and such exceptions are thrown by a constructor or method.

If a constructor or method declaration indicates thrown exceptions, javadoc creates the "Throws:" section in that declaration documentation. Here's the algorithm which javadoc uses to fill in that section:

1. For each `@throws` tag that is directly present on the declaration, add an entry to the section.
2. If this is a method declaration, then for those exceptions from that declaration's `throws` clause for which there were no `@throws` tags found in step 1, try to inherit `@throws` tags from the overridden methods found as per Method Comments Algorithm[^1].
3. For each of the remaining exceptions from the `throws` clause (i.e. the exceptions for which documentation could neither be found in step 1, nor inherited in step 2), add an entry that mentions the exception but has no description. 

The problem that JDK-8067757 is concerned with is that if an exception is documented using multiple `@throws` tags, only one of these tags will be inherited in step 2.

While fixing this issue I discovered an unpleasant interference with JDK-4947455[^2] and fixed it.

Looking ahead, JDK-6509045[^3] is about a similar problem that happens in step 1 in the presence of `{@inheritDoc}`. I'm also planning to fix JDK-6509045 in JDK 19.

[^1]: https://docs.oracle.com/en/java/javase/18/docs/specs/javadoc/doc-comment-spec.html#method-comments-algorithm
[^2]: https://bugs.openjdk.org/browse/JDK-4947455
[^3]: https://bugs.openjdk.org/browse/JDK-6509045

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8067757](https://bugs.openjdk.org/browse/JDK-8067757): Incorrect HTML generation for copied javadoc with multiple @throws tags

### Reviewers
 * [Jonathan Gibbons](https://openjdk.org/census#jjg) (@jonathan-gibbons - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/95/head:pull/95` \
`$ git checkout pull/95`

[...]

PR #116, opened 2022-07-06
6509045

pavelrappo:
This is the second of the two PRs that fix inheritance for multiple `@throws` tags for a particular exception.

The first PR (8067757, #95) fixed inheritance for multiple `@throws` tags for an exception referred to by a `throws` clause. This PR fixes inheritance for multiple `@throws` tags for an exception referred to by an `@throws` tag.

It is expected that if an overriding method has an `@throws` tag that has an `{@inheritDoc}` in its description, then the documentation will inherit all tags that describe that same exception in the overridden methods. For example, support the setup looks like this:

* Overrider
   ```
   @throws NullPointerException if this
   @throws NullPointerException if that
   ```

* Overridee
   ```
   @throws NullPointerException {@inheritDoc}
   ```
Then the overridee's documentation should look like this:
```
Throws:
NullPointerException - if this
NullPointerException - if that 
```

While logical and expected, such behavior is different from other use cases of `{@inheritDoc}`. In other use cases `{@inheritDoc}` expands locally: it replaces itself with a list of tree nodes. Such a replacement doesn't affect any tree node other than its parent.

`{@inheritDoc}` expands locally everywhere, such as in the main description, `@param`, and `@return` tags, but it is not enough for `@throws`, which sometimes needs to amend the **grandparent** node to replace a `@throws` node with multiple `@throws` nodes. Amending this behavior globally just for `@throws` would be impractical.

For now, it seems reasonable to fix this by overriding the `{@inheritDoc}` processing just for `@throws`. This PR gives `ThrowsTaglet` an ability to peek at `@throws` description to see if there are any `{@inheritDoc}` there.

There's one limitation though. Suppose `@throws` documentation is both amended and inhertied from multiple `@throws` like this:

* Overridder
```
@throws NullPointerException if this
@throws NullPointerException if that
```

* Overridee
```
@throws NullPointerException "{@inheritDoc}"
```

It's not clear what to do in such a case. If only one of either amendment (a `"` quote before and after) or multiple-inheritance is present, it's clear what to do. But not when both are present. For now, such a situation is detected and flagged as an error (a test is provided).

### DIFF
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
@@ -25,7 +25,6 @@
 
 package jdk.javadoc.internal.doclets.toolkit.taglets;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -34,7 +33,6 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
 
 import javax.lang.model.element.Element;
@@ -49,10 +47,8 @@ import com.sun.source.doctree.DocTree;
 import com.sun.source.doctree.ThrowsTree;
 
 import jdk.javadoc.doclet.Taglet.Location;
-import jdk.javadoc.internal.doclets.toolkit.BaseConfiguration;
 import jdk.javadoc.internal.doclets.toolkit.Content;
 import jdk.javadoc.internal.doclets.toolkit.util.DocFinder;
-import jdk.javadoc.internal.doclets.toolkit.util.Utils;
 
 /**
  * A taglet that processes {@link ThrowsTree}, which represents
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
@@ -25,6 +25,7 @@
 
 package jdk.javadoc.internal.doclets.toolkit.taglets;
 
+import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
```diff
@@ -175,13 +176,44 @@ private Content throwsTagsOutput(Map<ThrowsTree, ExecutableElement> throwsTags,
...
+                // 1. try to inherit
+                // 2. if tag.getDescription() adds to it, then:
+                //   2.1. if inherited none or one go skip, delegate to the old code
+                //   2.2. otherwise (inherited more than one tag) raise an error
+                // 3. otherwise (tag does not add to it), add all tags
+                var input = new DocFinder.Input(utils, e, this,
+                        new DocFinder.DocTreeInfo(tag, e), false, true);
+                var output = DocFinder.search(writer.configuration(), input);
+                if (output.tagList.size() <= 1) {
+                    // old code will doo just fine
```
jonathan-gibbons: typo: `doo`

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
```diff
@@ -175,13 +176,44 @@ private Content throwsTagsOutput(Map<ThrowsTree, ExecutableElement> throwsTags,
...
+                    tags.add(tag);
+                } else if (tag.getDescription().size() > 1) { // there's more inside @throws than just {@inheritDoc}
+                    // error (cannot do this reliably, probably is a programmatic error)
+                    // TODO: warn
```
jonathan-gibbons: Will the TODO addressed?

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
```diff
@@ -177,7 +177,8 @@ private Content throwsTagsOutput(Map<ThrowsTree, ExecutableElement> throwsTags,
                 result.add(writer.getThrowsHeader());
             }
 
-            final List<DocTree> tags = new ArrayList<>();
+            // basically, flatmap... @throws -> @throws*
+            final Map<DocTree, Element> tags = new LinkedHashMap<>(); //
```
jonathan-gibbons: empty '//'

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
```diff
@@ -221,6 +194,38 @@ private Content throwsTagsOutput(Map<ThrowsTree, ExecutableElement> throwsTags,
...
+            // 1. try to inherit
+            // 2. if tag.getDescription() adds to it, then:
+            //   2.1. if inherited none or one go skip, delegate to the old code
+            //   2.2. otherwise (inherited more than one tag) raise an error
+            // 3. otherwise (tag does not add to it), add all tags
+            var input = new DocFinder.Input(utils, e, this,
+                    new DocFinder.DocTreeInfo(tag, e), false, true);
+            var output = DocFinder.search(configuration, input);
+            if (output.tagList.size() <= 1) {
+                // old code will doo just fine
```
jonathan-gibbons: the typo got relocated

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/taglets/ThrowsTaglet.java
```diff
@@ -195,34 +195,34 @@ private Content throwsTagsOutput(Map<ThrowsTree, ExecutableElement> throwsTags,
     }
 
     private Map<ThrowsTree, Element> expand(ThrowsTree tag, Element e, Utils utils, BaseConfiguration configuration) {
+        // although we could use LinkedHashMap for a single mapping too, to emphasize
+        // that the order is important, a Map.of() would do just fine
         // basically, flatmap... @throws -> @throws*
         if (tag.getDescription().stream().noneMatch(d -> d.getKind() == DocTree.Kind.INHERIT_DOC)) {
             return Map.of(tag, e);
+        }
+        // @throws is the only tag where it currently makes sense
```
jonathan-gibbons: I don't think it always makes sense, so I would weaken this to _where it may currently make sense_ or similar.

It can make sense if the entire description is `{@inheritDoc}`
I don't think it makes sense when `{@inheritDoc}` is part of a longer description, with additional text and tags alongside `{@inheritDoc}`. I would be inclined to specify that as a format restriction for the use of `{@inherotDoc}` in `@throws`.

jonathan-gibbons (approved):
I like the general improvement of the abstractions in `ThrowsTaglet`.

I like reporting the error in the problem case.

PR #120, opened 2022-07-07
JDK-8289894: A NullPointerException thrown from guard expression

lahodaj:
Consider the following code:
```
public class SwitchGuard {
    public static void main(String... args) {
        test(null);
    }
    private static void test(Object o) {
        switch (o) {
            case null, String s when s.isEmpty() -> System.err.println("OK.");
            default -> {}
        }
    }
}
```

In the current specification, the guard should not be invoked when `o` is `null`, but javac will invoke it, for historical reasons.

Also, as opposed to JDK 18/JEP 420, `case null, <pattern>` is allowed for parenthesized pattern when the parenthesized pattern encloses (directly or via other parenthesized patterns) a type pattern.

The patch proposed here strives to fix both of these problems.

---------
### Progress

### Issue
 * [JDK-8289894](https://bugs.openjdk.org/browse/JDK-8289894): A NullPointerException thrown from guard expression

### Reviewers
 * [Vicente Romero](https://openjdk.org/census#vromero) (@vicente-romero-oracle - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/120/head:pull/120` \
`$ git checkout pull/120`

`$ git checkout pull/120` \
`$ git pull https://git.openjdk.org/jdk19 pull/120/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 120`

`$ git pr show -t 120`

</details>

<a href="https://git.openjdk.org/jdk19/pull/120.diff">https://git.openjdk.org/jdk19/pull/120.diff</a>

</details>

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Check.java
@@ -4348,6 +4348,9 @@ public class Check {
                     wasDefault = true;
                 } else {
                     JCPattern pat = ((JCPatternCaseLabel) label).pat;
+                    while (pat instanceof JCParenthesizedPattern parenthesized) {
+                        pat = parenthesized.pattern;
+                    }
                     boolean isTypePattern = pat.hasTag(BINDINGPATTERN);
                     if (wasPattern || wasConstant || wasDefault ||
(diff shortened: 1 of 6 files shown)


biboudis:
LGTM 🎉

PR #130, opened 2022-07-11
8289954: C2: Assert failed in PhaseCFG::verify() after JDK-8183390

pfustc:
Fuzzer tests report an assertion failure issue in C2 global code motion
phase. Git bisection shows the problem starts after our fix of post loop
vectorization (JDK-8183390). After some narrowing down work, we find it
is caused by below change in that patch.

```
@@ -422,14 +404,7 @@
       cl->mark_passed_slp();
     }
     cl->mark_was_slp();
-    if (cl->is_main_loop()) {
-      cl->set_slp_max_unroll(local_loop_unroll_factor);
-    } else if (post_loop_allowed) {
-      if (!small_basic_type) {
-        // avoid replication context for small basic types in programmable masked loops
-        cl->set_slp_max_unroll(local_loop_unroll_factor);
-      }
-    }
+    cl->set_slp_max_unroll(local_loop_unroll_factor);
   }
 }
```

This change is in function `SuperWord::unrolling_analysis()`. AFAIK, it
helps find a loop's max unroll count via some analysis. In the original
code, we have loop type checks and the slp max unroll value is set for
only some types of loops. But in JDK-8183390, the check was removed by
mistake. In my current understanding, the slp max unroll value applies
to slp candidate loops only - either main loops or RCE'd post loops -
so that check shouldn't be removed. After restoring it we don't see the
assertion failure any more.

The new jtreg created in this patch can reproduce the failed assertion,
which checks `def_block->dominates(block)` - the domination relationship
of two blocks. But in the case, I found the blocks are in an unreachable
inner loop, which I think ought to be optimized away in some previous C2
phases. As I'm not quite familiar with the C2's global code motion, so
far I still don't understand how slp max unroll count eventually causes
that problem. This patch just restores the if condition which I removed
incorrectly in JDK-8183390. But I still suspect that there is another
hidden bug exists in C2. I would be glad if any reviewers can give me
some guidance or suggestions.

Tested hotspot::hotspot_all_no_apps, jdk::tier1~3 and langtools::tier1.

---------
### Progress

### Issue
 * [JDK-8289954](https://bugs.openjdk.org/browse/JDK-8289954): C2: Assert failed in PhaseCFG::verify() after JDK-8183390

### Reviewers
 * [Tobias Hartmann](https://openjdk.org/census#thartmann) (@TobiHartmann - **Reviewer**)
 * [Roland Westrelin](https://openjdk.org/census#roland) (@rwestrel - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/130/head:pull/130` \
`$ git checkout pull/130`

`$ git checkout pull/130` \
`$ git pull https://git.openjdk.org/jdk19 pull/130/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 130`

`$ git pr show -t 130`

</details>

<a href="https://git.openjdk.org/jdk19/pull/130.diff">https://git.openjdk.org/jdk19/pull/130.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/superword.cpp
+++ b/src/hotspot/share/opto/superword.cpp
@@ -414,7 +414,9 @@ void SuperWord::unrolling_analysis(int &local_loop_unroll_factor) {
       cl->mark_passed_slp();
     }
     cl->mark_was_slp();
-    cl->set_slp_max_unroll(local_loop_unroll_factor);
+    if (cl->is_main_loop() || cl->is_rce_post_loop()) {
+      cl->set_slp_max_unroll(local_loop_unroll_factor);
+    }
   }
 }
 
--- a/test/hotspot/jtreg/compiler/loopopts/TestUnreachableInnerLoop.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestUnreachableInnerLoop.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2022, Arm Limited. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8289954
+ * @summary C2: Assert failed in PhaseCFG::verify() after JDK-8183390
+ *
+ * @run main/othervm -Xcomp -Xbatch
+ *      -XX:CompileOnly=compiler/loopopts/TestUnreachableInnerLoop
+ *      compiler.loopopts.TestUnreachableInnerLoop
+ */
+
+package compiler.loopopts;
+
+public class TestUnreachableInnerLoop {
+
+    public static int field = 0;
+    public static int arr[] = new int[500];
+
+    public static void fun() {
+        for (int elem : arr) {
+            int x = 1, y = 2, z = 3;
+            int i, j;
+
+            // This is a good loop
+            for (i = 2; i < 63; i++) {
+                arr[i] = arr[i] + 3592870;
+            }
+
+            // The inner loop looks quite complex but it's unreachable
+            // code as loop condition "k < 2" never satisfies
+            for (j = 3; j < 63; j++) {
+                for (int k = j; k < 2; k++) {
+                    arr[j] <<= k;
+                    try {
+                        x = k / i;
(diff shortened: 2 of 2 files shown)


vnkozlov (approved):
Testing passed.

vnkozlov:
> @vnkozlov Thanks for looking at this. I think a 2nd review is required, right?

yes

pfustc:
@dean-long Do you have any comments or suggestions on this? The failure was reported from your fuzzer test.

dean-long:
@pfustc Sorry, I'm not enough of an expert on SuperWord to review the fix.  The test was generated automatically by Java Fuzzer.

pfustc:
> @pfustc Sorry, I'm not enough of an expert on SuperWord to review the fix. The test was generated automatically by Java Fuzzer.

May I ask how do you generate and run the Fuzzer tests? Is there any instructions we can follow? Recently we see a couple of SuperWord issues reported by corner cases which are generated by the Fuzzer.

dean-long:
@pfustc take a look at https://github.com/AzulSystems/JavaFuzzer.

pfustc:
> @pfustc take a look at https://github.com/AzulSystems/JavaFuzzer.

Thanks Dean. I will investigate that project.

pfustc:
@rwestrel @TobiHartmann Would you like to review this fix for jdk19? The RDP1 will end in one week.

rwestrel:
To have this fixed in jdk 19, you need to open a PR againsts jdk 19.
Isn't SuperWord::unrolling_analysis() only called for main and rce post loops anyway? What loop type doesn't the extra check exclude in the case of your test?

TobiHartmann (approved):
Looks reasonable to me.

TobiHartmann:
> To have this fixed in jdk 19, you need to open a PR againsts jdk 19.

But this is a PR against JDK 19, right?

rwestrel:
> To have this fixed in jdk 19, you need to open a PR againsts jdk 19.

Sorry I missed this was indeed against jdk 19.

rwestrel:
> We probably need another fix to avoid this. But to reduce risks, I still propose we just restore the incorrectly updated code in this PR for jdk 19 and do complete fix in jdk 20.

That's reasonable.

rwestrel (approved):
Looks good to me.

RealFYang:
Sorry, but seems that the same assertion failure is still happening when running the newly added test case with fastdebug build on linux-riscv64 platform. And I have attached the hs_err and reply files on the JBS issue. Please take another look.

pfustc:
Thanks @RealFYang for the information. I'm still investigating this in jdk 20 but so far I haven't got a clear clue. Just find if I ban `SuperWord::unrolling_analysis()` for normal loops, specifically with below change, that assertion fails again.
```
diff --git a/src/hotspot/share/opto/superword.cpp b/src/hotspot/share/opto/superword.cpp
index ef66840628f..997fae367b1 100644
--- a/src/hotspot/share/opto/superword.cpp
+++ b/src/hotspot/share/opto/superword.cpp
@@ -217,6 +217,7 @@ void SuperWord::unrolling_analysis(int &local_loop_unroll_factor) {
   int *ignored_loop_nodes = NEW_RESOURCE_ARRAY(int, ignored_size);
   Node_Stack nstack((int)ignored_size);
   CountedLoopNode *cl = lpt()->_head->as_CountedLoop();
+  if (cl->is_normal_loop()) return;
   Node *cl_exit = cl->loopexit_or_null();
   int rpo_idx = _post_block.length();
```

And I saw [JDK-8275330](https://github.com/openjdk/jdk/pull/6429) by Roland fixed the same assertion failure before. And the test case which causes the failure looks similar with this one - there is an inner dead loop which fails to be optimized away by C2.

Hi @rwestrel , may I ask if you have any ideas or hints about this?

TobiHartmann:
@pfustc Please file a new bug for the remaining issue. Thanks!

pfustc:
@RealFYang, I have created a new JBS: https://bugs.openjdk.org/browse/JDK-8291025 and attached your hs_err_* file. Feel free to edit it if you have something to add.

BTW: It may be helpful if you could provide the output with VM option `-XX:+TraceSuperWordLoopUnrollAnalysis`.

PR #139, opened 2022-07-13
8290252: Add TEST.properties to java/nio/channels/FileChannel and move tests out of largeMemory sub-dir

bplb:
Attempt to make `FIleChannel` tests not run concurrently with each other.


jaikiran (approved):

AlanBateman:
> just curious as to why TEST.properties was added in test directory java/nio/channels/FileChannel rather than amending the TEST.ROOT java/nio/channels/FileChannel/largeMemory entry to java/nio/channels/FileChannel to get exclusive non current execution ?

Personally I think it is easier to find when it's the same directory has the tests. There are currently >90 TEST.properties files in the test/jdk tree, but you are right that it could work both ways.

PR #140, opened 2022-07-13
8290250: Shenandoah: disable Loom for iu mode

zhengyu123:
Please review this trivial patch to disable Loom for Shenandoah iu mode (an experimental feature).

Test:
- [x] hotspot_gc_shenandoah
- [x] jdk_loom with Shenandoah + iu
- [x] hotspot_loom with Shenandoah + iu

---------
### Progress

### Issue
 * [JDK-8290250](https://bugs.openjdk.org/browse/JDK-8290250): Shenandoah: disable Loom for iu mode

### Reviewers
 * [Roman Kennke](https://openjdk.org/census#rkennke) (@rkennke - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/140/head:pull/140` \
`$ git checkout pull/140`

`$ git checkout pull/140` \
`$ git pull https://git.openjdk.org/jdk19 pull/140/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 140`

`$ git pr show -t 140`

</details>

<a href="https://git.openjdk.org/jdk19/pull/140.diff">https://git.openjdk.org/jdk19/pull/140.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/shenandoah/mode/shenandoahIUMode.cpp
+++ b/src/hotspot/share/gc/shenandoah/mode/shenandoahIUMode.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2021, Red Hat, Inc. All rights reserved.
+ * Copyright (c) 2020, 2022, Red Hat, Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -50,6 +50,9 @@ void ShenandoahIUMode::initialize_flags() const {
   if (FLAG_IS_DEFAULT(ShenandoahSATBBarrier)) {
     FLAG_SET_DEFAULT(ShenandoahSATBBarrier, false);
   }
+
+  // Disable Loom
+  SHENANDOAH_ERGO_DISABLE_FLAG(VMContinuations);
 
   SHENANDOAH_ERGO_ENABLE_FLAG(ExplicitGCInvokesConcurrent);
   SHENANDOAH_ERGO_ENABLE_FLAG(ShenandoahImplicitGCInvokesConcurrent);


fisk:
Speaking of IU mode, how does Shenandoah IU mode deal with barrier elision on newly allocated objects? It was only valid in CMS because the entire young generation was traversed when terminating the concurrent marking, so that unvisited pointers could be found.

rkennke (approved):
Yes, looks good. Thank you.

zhengyu123:
> > > Speaking of IU mode, how does Shenandoah IU mode deal with barrier elision on newly allocated objects? It was only valid in CMS because the entire young generation was traversed when terminating the concurrent marking, so that unvisited pointers could be found.
> > 
> > 
> > Shenandoah SATB/IU both elide barrier on newly allocated objects. SATB barrier should guarantee they can only refer marked objects. Do I miss anything?
> 
> Yes, I believe so. For a SATB collector, it is okay to elide the barrier on newly allocated objects, because they were not part of the snapshot-of-the-beginning, and all new objects are implicitly alive and don't need to be visited.
> 
> However, with an IU scheme, that property does not translate.
> 
> Consider that you have a particular object graph when marking starts, then one mutator loads an object from the graph, and clears the field such that it is only reachable from the roots of said murator thread. Then the mutator writes the object to a newly allocated object without barriers and discards the root. Now, the reference is only reachable from the newly allocated object.
> 
> So for this elision to be valid with IU, the GC has to visit newly allocated objects again before terminating, which e.g. CMS indeed did, by doing a young collection in the safepoint that terminated old marking, but I don't know that Shenandoah does anything like that.

Unfortunately, IU problem is beyond the barrier elision. Even I don't elide the barrier for IU, I still see failures.

fisk:
> > > > Speaking of IU mode, how does Shenandoah IU mode deal with barrier elision on newly allocated objects? It was only valid in CMS because the entire young generation was traversed when terminating the concurrent marking, so that unvisited pointers could be found.
> 
> > > 
> 
> > > 
> 
> > > Shenandoah SATB/IU both elide barrier on newly allocated objects. SATB barrier should guarantee they can only refer marked objects. Do I miss anything?
> 
> > 
> 
> > Yes, I believe so. For a SATB collector, it is okay to elide the barrier on newly allocated objects, because they were not part of the snapshot-of-the-beginning, and all new objects are implicitly alive and don't need to be visited.
> 
> > 
> 
> > However, with an IU scheme, that property does not translate.
> 
> > 
> 
> > Consider that you have a particular object graph when marking starts, then one mutator loads an object from the graph, and clears the field such that it is only reachable from the roots of said murator thread. Then the mutator writes the object to a newly allocated object without barriers and discards the root. Now, the reference is only reachable from the newly allocated object.
> 
> > 
> 
> > So for this elision to be valid with IU, the GC has to visit newly allocated objects again before terminating, which e.g. CMS indeed did, by doing a young collection in the safepoint that terminated old marking, but I don't know that Shenandoah does anything like that.
> 
> 
> 
> Unfortunately, IU problem is beyond the barrier elision. Even I don't elide the barrier for IU, I still see failures.

Right. Loom assumes that write barriers are not needed on newly allocated objects, so that it can copy oops from the stack to the heap. Shenandoah with IU mode is the only collector where that is not true, due to the reason I outlined. What I'm saying isn't that turning off barrier elision in IU mode will fix your current loom crash. I'm saying that even without loom, the barrier elision is invalid in general. I thought I'd let you know, in case you want Shenandoah with IU mode to work reliably at all.

rkennke:
> > > Speaking of IU mode, how does Shenandoah IU mode deal with barrier elision on newly allocated objects? It was only valid in CMS because the entire young generation was traversed when terminating the concurrent marking, so that unvisited pointers could be found.
> > 
> > 
> > Shenandoah SATB/IU both elide barrier on newly allocated objects. SATB barrier should guarantee they can only refer marked objects. Do I miss anything?
> 
> Yes, I believe so. For a SATB collector, it is okay to elide the barrier on newly allocated objects, because they were not part of the snapshot-of-the-beginning, and all new objects are implicitly alive and don't need to be visited.
> 
> However, with an IU scheme, that property does not translate.
> 
> Consider that you have a particular object graph when marking starts, then one mutator loads an object from the graph, and clears the field such that it is only reachable from the roots of said murator thread. Then the mutator writes the object to a newly allocated object without barriers and discards the root. Now, the reference is only reachable from the newly allocated object.
> 
> So for this elision to be valid with IU, the GC has to visit newly allocated objects again before terminating, which e.g. CMS indeed did, by doing a young collection in the safepoint that terminated old marking, but I don't know that Shenandoah does anything like that.

IIRC, in Shenandoah's IU mode we treat new objects as implicitely alive, much like we do with SATB mode. Therefore it should be good to elide the barriers there. I know that this is not the classical I-U scheme that I believe CMS followed. There has been a paper (I don't remember the name) that showed that it's orthogonal issues whether we follow new or old references and whether or not we treat new objects alive or not. Treating new objects as live has the advantage that it solves the termination problem.

zhengyu123:
> _Mailing list message from [Erik Osterlund](mailto:erik.osterlund@oracle.com) on [shenandoah-dev](mailto:shenandoah-dev@mail.openjdk.org):_
> 
> Hi Zhengyu,
> 
> I suppose requires_barriers needs to return true when marking is active with IU, as you need barriers even on newly allocated objects then, as I was discussing.
> 
> /Erik

Tried that, still not working.
I can not see how it can be Loom specific? but seems it's only exposed by `Skynet`

Thanks,

-Zhengyu

PR #152, opened 2022-07-22
8290867: Race freeing remembered set segments

tschatzl:
Hi all,

  please review this fix for a crash due to a race in remembered set segment deallocation. Here is the description (provided by chaeubl as reported):

 - Thread A executes `G1SegmentedArray::create_new_segment` and tries to pop an element from the `_free_segment_list`. For that, thread A executes `LockFreeStack::pop()`
- Thread A reads `LockFreeStack::top()`
- Thread B executes `LockFreeStack::pop()`, also reads `LockFreeStack::top()` and pops that element from the stack
- Thread B executes `Atomic::cmpxchg(&_first, prev, next);` in `G1SegmentedArray::create_new_segment` but it fails because another thread already registered a different segment
- Thread B calls `G1SegmentedArraySegment::delete_segment` and frees the value
- Thread A tries to access `top()->next` in `LockFreeStack::pop()`, which causes a segfault because `top()` was freed by thread B 

The fix is to delay the deletion of that memory segment until all readers (i.e. in `G1SegmentedArrayFreeList::get` calling `_list.pop()`) drop the references to that memory segment. The readers are already guarded by a `CriticalSection`.

Testing: tier1-5 running, reproducer that adds extra delays that significantly delays to widen the opportunity this race can occur passes on BigRAMTester (otherwise crashes in a few seconds)

Thanks,
  Thomas

---------
### Progress

### Issue
 * [JDK-8290867](https://bugs.openjdk.org/browse/JDK-8290867): Race freeing remembered set segments

### Reviewers
 * [Kim Barrett](https://openjdk.org/census#kbarrett) (@kimbarrett - **Reviewer**) ⚠️ Review applies to [2ec1d209](https://git.openjdk.org/jdk19/pull/152/files/2ec1d209956f3f414473abc74cbafc7195a70be6)
 * [Sangheon Kim](https://openjdk.org/census#sangheki) (@sangheon - **Reviewer**) ⚠️ Review applies to [2ec1d209](https://git.openjdk.org/jdk19/pull/152/files/2ec1d209956f3f414473abc74cbafc7195a70be6)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/152/head:pull/152` \
`$ git checkout pull/152`

`$ git checkout pull/152` \
`$ git pull https://git.openjdk.org/jdk19 pull/152/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 152`

`$ git pr show -t 152`

</details>

<a href="https://git.openjdk.org/jdk19/pull/152.diff">https://git.openjdk.org/jdk19/pull/152.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/g1/g1SegmentedArray.cpp
+++ b/src/hotspot/share/gc/g1/g1SegmentedArray.cpp
@@ -27,6 +27,7 @@
 #include "gc/g1/g1SegmentedArray.inline.hpp"
 #include "memory/allocation.hpp"
 #include "runtime/atomic.hpp"
+#include "runtime/vmOperations.hpp"
 #include "utilities/globalCounter.inline.hpp"
 
 G1SegmentedArraySegment::G1SegmentedArraySegment(uint slot_size, uint num_slots, G1SegmentedArraySegment* next, MEMFLAGS flag) :
@@ -48,8 +49,11 @@ G1SegmentedArraySegment* G1SegmentedArraySegment::create_segment(uint slot_size,
 }
 
 void G1SegmentedArraySegment::delete_segment(G1SegmentedArraySegment* segment) {
-  // Wait for concurrent readers of the segment to exit before freeing.
-  GlobalCounter::write_synchronize();
+  // Wait for concurrent readers of the segment to exit before freeing; but only if the VM
+  // isn't exiting.
+  if (!VM_Exit::vm_exited()) {
+    GlobalCounter::write_synchronize();
+  }
   segment->~G1SegmentedArraySegment();
   FREE_C_HEAP_ARRAY(_mem_flag, segment);
 }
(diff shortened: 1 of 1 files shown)


kimbarrett (approved):
Looks good.

sangheon (approved):
Looks good.

tschatzl:
Testing showed an issue with VM shutdown: VM shutdown calls the the destructor for the (global) free list pool, which in turn frees all segments; however at that time the VM is not in a state where `GlobalCounter` works (threads are detached, no valid threads list) and crashes.
The last commit provides a workaround for that issue, not doing the global synchronization when shutting down. That is safe because the process is single-threaded in the `VM_Exit` operation anyway.
This is a workaround for this issue that needs to be cleaned up, but it's too late imho in the release to find a better solution for this issue.

Passes tier1-5 with no crashes.

tschatzl:
Thanks @kimbarrett @sangheon for your reviews
/integrate

kimbarrett:
> Testing showed an issue with VM shutdown: VM shutdown calls the the destructor for the (global) free list pool, which in turn frees all segments; however at that time the VM is not in a state where `GlobalCounter` works (threads are detached, no valid threads list) and crashes. The last commit provides a workaround for that issue, not doing the global synchronization when shutting down. That is safe because the process is single-threaded in the `VM_Exit` operation anyway. This is a workaround for this issue that needs to be cleaned up, but it's too late imho in the release to find a better solution for this issue.

For the record, we discussed this before Thomas pushed, and I agree with his assessment.  The workaround is simple and safe, and doing something more principled might not fit in the current release schedule.

PR #153, opened 2022-07-22
8287794: Reverse*VNode::Identity problem

jatin-bhateja:
Hi All,

- This bug fix patch fixes a missing case during reverse[bits|bytes] identity transformation.
- Unlike AARCH64(SVE),  X86(AVX512) ISA has no direct instruction to reverse[bits|bytes] of a vector lane hence a predicated operation is supported through blend instruction.
- New IR framework based tests have been added for transforms relevant to AVX2, AVX512 and SVE.

Kindly review and share your feedback.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8287794](https://bugs.openjdk.org/browse/JDK-8287794): Reverse*VNode::Identity problem

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jdk19 pull/153/head:pull/153` \
`$ git checkout pull/153`

`$ git checkout pull/153` \
`$ git pull https://git.openjdk.org/jdk19 pull/153/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 153`

`$ git pr show -t 153`

</details>

<a href="https://git.openjdk.org/jdk19/pull/153.diff">https://git.openjdk.org/jdk19/pull/153.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/vectornode.cpp
+++ b/src/hotspot/share/opto/vectornode.cpp
@@ -1725,7 +1725,9 @@ Node* ReverseBytesVNode::Identity(PhaseGVN* phase) {
   }
   // ReverseBytesV (ReverseBytesV X , MASK) , MASK =>  X
   if (in(1)->Opcode() == Op_ReverseBytesV) {
-    if (is_predicated_vector() && in(1)->is_predicated_vector() && in(2) == in(1)->in(2)) {
+    if (is_predicated_vector() ^ in(1)->is_predicated_vector()) {
+      return this;
+    } else if (is_predicated_vector() && in(1)->is_predicated_vector() && in(2) == in(1)->in(2)) {
       return in(1)->in(1);
     } else {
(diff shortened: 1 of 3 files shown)


vnkozlov (changes requested):
This is P4 bug and we are in phase 2 for JDK 19 release. The bug's fix version is JDK 20.
Please, create new PR based on latest JDK and not JDK 19. We can later backport it into JDK 19 update release.

dean-long (changes requested):
The "else if" clause still returns the same value as the "else" clause, so why is the "else if" part needed?  This change doesn't seem to address the question asked by the submitter of the bug:
>  Seems to me the first condition checks that MASKs are the same in both nodes. But if they are not, we are falling to "else" branch, where we do the same transformation anyway. So, there might be a bug lurking there when MASKs are different.

jatin-bhateja:
Closing this PR. Will create one against JDK mainline with resolved review comments.

